PR #3, opened 2016-05-05
fixed invalid sytax in utils.py, line 201

jmwolff3:


mdipierro:
this PR is about a very old version of web3py that no longer exists.

PR #17, opened 2019-05-30
update dockerfile using ubuntu and include web3py test

sugizo:
https://groups.google.com/d/msg/web2py/saav4hQdbSU/x7T1OES6BwAJ

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,31 @@
-# Change anything you need to in this file, including the base image.
-FROM python:3.7
+# please change :
+# password.txt with your desire filename
+# password_admin with your desire password admin
+# 8000 with your available port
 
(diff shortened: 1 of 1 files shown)


mdipierro:
Thanks for your work. Is there a lighter image we can use instead of ubuntu:latest?

sugizo:
tested work only with debian derivatives : ubuntu n debian, will try with alpine, want to create dockerfile separately or just one ?
e.g. for separately
debian/dockerfile
alpine/dockerfile
?

sugizo:
tested related with alpine Dockerfile not work due to /bin/sh: git not found, still find the solution
detail steps on 
https://groups.google.com/d/msg/web2py/saav4hQdbSU/DPIHX4ekCAAJ

PR #37, opened 2019-08-31
Create missing _scaffold/databases folder

MBfromOK:
fixes issue with the _scaffold/databases folder not existing


MBfromOK:
Realized this should be handled by code instead of creating an empy folder.. will make pull request for that.

PR #38, opened 2019-08-31
Source: Various startup fixes.. please review, not all are necessarily best use / case.

MBfromOK:
**Added:** /apps/\_\_init\_\_.py - fixes source launch issue
**Added:** assets folder from pip package to source install

# py4web-start
**Changed:** py4web-start to py4web-start.py  - Starting py4web-start from the command line (in windows) with any of the following fails:
- ./py4web-start apps
- .\py4web-start apps
- /py4web-start apps
- /py4web-start apps

with the message:

> '<command above\>' is not recognized as an internal or external command,
operable program or batch file.

# /apps/examples/\_\_init\_\_.py

Loading py4web from source the examples module fails to load due to a missing databases folder.

Focusing strictly on this app, felt the \_\_init\_\_.py  was the most appropriate place to add code to create the databases folder if it is missing.

## Added:
```
if not os.path.exists(os.path.join(os.path.dirname(__file__), 'databases')):
    try:
        os.makedirs(os.path.join(os.path.dirname(__file__), 'databases'))
    except FileExistsError:
        pass
```

# /apps/_scaffold/models.py

Loading py4web from source the _scaffold app fails to load due to a missing databases folder.

Focusing strictly on this app, felt the models.py  was the most appropriate place to add code to create the databases folder if it is missing.

## Added: 
```
if not os.path.exists(settings.DB_FOLDER):
    try:
        os.makedirs(settings.DB_FOLDER)
    except FileExistsError:
        pass
```


mdipierro:
Your tests and suggestions are very valuable. Since I had to pick some and not others I did it in a separate commit and acknowledged your contribution in the commit message and the main page. I will therefore reject this PR.

There is one outstading issue: the assets. You propose including them. I think they should not be but I see the added complication of running "make assets" which is not easy on windows.

It would be great if you could write a python function that crates the assets using os.walk and zipfile. Then we would not need make assets or committing the zip files.

MBfromOK:
> There is one outstading issue: the assets. You propose including them. I think they should not be but I see the added complication of running "make assets" which is not easy on windows.

I agree, my realization was that they shouldn't statically be included, but should be downloaded / verified from source / github if missing.

>It would be great if you could write a python function that crates the assets using os.walk and zipfile. Then we would not need make assets or committing the zip files.

I will take a look at this, I am a new python programmer and haven't even touched os.walk, zip, or assets yet (other than to find the ones included in the pip install and move them over).

PR #79, opened 2019-11-01
py4web-start --> py4web-start.py in docs

lucadealfaro:
Small fix to documentation.


MBfromOK:
@lucadealfaro @mdipierro 

Will using this syntax change will break IDE integrated scripting?

Launching the package version installed in pycharm with pip I run:

`python -m py4web-start apps --address localhost:8080`

or github clone:

`python py4web-start.py apps --address localhost:8080`

PR #90, opened 2019-12-02
Work on form validation

lucadealfaro:
This is not ready to be accepted, because the problem is that form.vars are NOT of the appropriate type, as they used to be in web2py.  That is, if I define a field

Field('product_quantity', 'integere')

then 

form.vars['product_quantity'] is NOT an integer; it is still a string.  I am wondering whether this difference is intentional.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -137,7 +137,7 @@ class Form(object):
                  keep_values=False,
                  form_name=False,
                  hidden=None,
-                 before_validate=None):
+                 validation=None):
 
         if isinstance(table, list):
             dbio = False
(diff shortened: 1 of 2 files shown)


py4web/utils/form.py
```diff
@@ -202,6 +200,8 @@ def __init__(self,
                             self.vars[field.name] = value
                             if error:
                                 self.errors[field.name] = error
+                    if len(self.errors) == 0 and validation:
+                        self.errors = validation(self)
```
mdipierro: No need to return errors given that validate can manipulate self.errors. Will change it after merging.

Issue #97, opened 2019-12-26
Git Merge: 12/26/2019 - _scaffold Error


MBfromOK:
Just merged git repo with my float using script to test #70  and I got the following error while trying to launch Py4Web:

Traceback (most recent call last):
  File "D:\Dev\dashproj\dashproj\py4web\py4web\core.py", line 825, in import_app
    module_name, init
  File "<frozen importlib._bootstrap_external>", line 407, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 907, in load_module
  File "<frozen importlib._bootstrap_external>", line 732, in load_module
  File "<frozen importlib._bootstrap>", line 265, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "apps\_scaffold\__init__.py", line 7, in <module>
    from . models import db
  File "apps\_scaffold\models.py", line 5, in <module>
    from .common import db, Field
  File "apps\_scaffold\common.py", line 102, in <module>
    from celery import Celery
ModuleNotFoundError: No module named 'celery'

Py4Web still suceeds in running Bottle and hosting.

MBfromOK:
FWIW: I overwrote everything I had with changes from Master

mdipierro:
sorry. Try now.

MBfromOK:
Tried again, still unable to test as the dashboard won't load.

Most recent attempt nearly mirrors #99 returning error:

Traceback (most recent call last):
  File "D:\Dev\dashproj\dashproj\py4web\py4web\core.py", line 826, in import_app
    module_name, init
  File "<frozen importlib._bootstrap_external>", line 407, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 907, in load_module
  File "<frozen importlib._bootstrap_external>", line 732, in load_module
  File "<frozen importlib._bootstrap>", line 265, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
[...]

PR #98, opened 2019-12-26
(Temporary) Patch for Celery Error: #97

MBfromOK:
Enclosed Celery load in try block to report missing module instead of returning a lengthy Traceback error. Issue Report #97


mdipierro:
Sorry. Better to just default USE_CELERY to false

PR #114, opened 2020-02-14
Luca signed urls

lucadealfaro:
This implements signed URLs.  Example of use: 

@action('testsig', method=['GET'])
@action.uses('testsig.html', session)
def testsig():
    """Test for URL signer"""
    return dict(myurl=URL('testurl', signer=signed_url))

@action('testurl', method=['GET'])
@action.uses('testurl.html', signed_url)
def testurl():
    return dict()

The one ugly thing is that one must remember to use @action.uses(session) when using the URL helper, as a key may need to be stored in the session if not already there.  Alternatively, one could ensure that sessions always contained such a key?  But that's also ugly.  What's the best way?


py4web/utils/factories.py
```diff
@@ -17,7 +17,7 @@ def post(self, path=None, template=None):
         return self._action_maker('POST', path, template)
 
     def put(self, path=None, template=None):
-        return self._action_maker('PUT', template, requires_login)
+        return self._action_maker('PUT', path, template)
```
mdipierro: oops

PR #116, opened 2020-02-14
Fixed some small issues with url signer

lucadealfaro:
I fixed some small issues with the code. 

Still, I don't see the need for a separate URLVerifier class; I would have simply made URLSigner a subclass of Fixture.  This would simplify the code and allow one to write 

@action.uses(url_signer)

rather than 

@action.uses(url_singer.verify())

I think in py4web spirit of keeping it small, if someone does not like our url singer, they could simply subclass it or provide their own, so I don't see the need for one additional class.

### DIFF
--- a/py4web/utils/url_signer.py
+++ b/py4web/utils/url_signer.py
@@ -10,66 +10,64 @@ class URLVerifier(Fixture):
      to action.uses() to check for the validity of signatures, and the
      sign() method can be used to sign a URL.  If an object of this class
      is passed to the URL helper, it can be used to sign a URL."""
-    def __init__(self, url_signer, query_keys):
+    def __init__(self, url_signer):
         if url_signer.session:
             self.__prerequisites__ = [url_signer.session]
         self.url_signer = url_signer
-        self.query_keys = query_keys
 
     def on_request(self):
         """Checks the request's signature"""
-        vars = dict(request.query)
         # extra and remove the signature from the query
         signature = request.query.get("_signature")
         if signature is None:
             abort(403)
         del request.query["_signature"]
-        # verify all query_keys or only those specified
-        query = {key: value for key, value in request.query.items()
-                 if self.query_keys is None or key in self.query_keys}
-        if signature != self.url_signer._sign(request.fullpath, query):
+        # Verifies the query keys.
+        if signature != self.url_signer._sign(request.fullpath, request.query):
             abort(403)
 
 
-class URLSigner:
-    def __init__(self, session=None, key=None, salt=b"", sign_query=False):
+class URLSigner(object):
+    def __init__(self, session=None, key=None, salt=b"", variables_to_sign=None):
         """
         you can provde a key or a session to sign the URL
         if none provided will use the global Session.SECRET
         salt is some salt that can be used in signing if desired.
-        if sign_query is set to false, the query string will also be signed
+        variables_to_sign is a list of variables to be included in the signature.
         """
         super().__init__(
         )  # Yes, I know that this currently doesn't do anything.
         self.session = session
         self.key = key or Session.SECRET
         self.salt = salt
-        self.sign_query = sign_query
+        self.variables_to_sign = variables_to_sign or []
+        assert '_signature' not in self.variables_to_sign
 
     def _get_key(self):
         """Gets the signing key, creating it if necessary."""
         if not self.session:
             key = self.key
         else:
-            self.session.get("_signature_key")
+            key = self.session.get("_signature_key")
             if key is None:
                 key = str(uuid.uuid4())
                 self.session["_signature_key"] = key
         return key
 
-    def _sign(self, url, query):
+    def _sign(self, url, vars):
         """Signs the URL"""
         h = hashlib.sha256(self.salt)
         h.update(url.encode('utf8'))  # Is utf8 the right encoding?
-        if self.sign_query:
-            for key in sorted(query):
-                h.update(('%s=%r' % (key, query[key])).encode('utf8'))
+        # Adds the variables that need to be signed.
+        for key in self.variables_to_sign:
+            h.update(('%s=%r' % (key, vars[key])).encode('utf8'))
         h.update(self._get_key().encode('utf8'))
         return h.hexdigest()
 
     def sign_vars(self, url, vars):
+        """Signs a URL, adding to vars (the variables of the URL) a signature."""
         vars["_signature"] = self._sign(url, vars)
 
-    def verify(self, query_keys=None):
+    def verify(self):
         """returns a fixture that verifies the URL and optionally the query_keys"""
-        return URLVerifier(self, query_keys)
+        return URLVerifier(self)


py4web/utils/url_signer.py
```diff
@@ -10,66 +10,64 @@ class URLVerifier(Fixture):
      to action.uses() to check for the validity of signatures, and the
      sign() method can be used to sign a URL.  If an object of this class
      is passed to the URL helper, it can be used to sign a URL."""
-    def __init__(self, url_signer, query_keys):
```
mdipierro: The reason for this was the following:

<a onclick="axios.get('[[=URL('xyz', vars=dict(x=1, y=2), signer=signer)]]'+'&z='+windows.data.z)">click me</a>

Here x,y are signed. z is not because generated in JS. How do we communicate this to the action? Should we not allow this? Maybe not, but why not?

py4web/utils/url_signer.py
```diff
@@ -10,66 +10,64 @@ class URLVerifier(Fixture):
...
+        self.variables_to_sign = variables_to_sign or []
+        assert '_signature' not in self.variables_to_sign
 
     def _get_key(self):
         """Gets the signing key, creating it if necessary."""
         if not self.session:
             key = self.key
         else:
-            self.session.get("_signature_key")
+            key = self.session.get("_signature_key")
```
mdipierro: oops

py4web/utils/url_signer.py
```diff
@@ -10,66 +10,64 @@ class URLVerifier(Fixture):
...
 
-    def _sign(self, url, query):
+    def _sign(self, url, vars):
         """Signs the URL"""
         h = hashlib.sha256(self.salt)
         h.update(url.encode('utf8'))  # Is utf8 the right encoding?
-        if self.sign_query:
-            for key in sorted(query):
-                h.update(('%s=%r' % (key, query[key])).encode('utf8'))
+        # Adds the variables that need to be signed.
```
mdipierro: I see. You want one signed for action. If this is the intention, I may ok with it.

py4web/utils/url_signer.py
lucadealfaro: Yes, and you can still do the same.   The list of keys to be signed is part of the url_signer, that is passed as an argument, so there is no need to give separately the list of query_keys to be signed.

py4web/utils/url_signer.py
lucadealfaro: As you like for the two classes.  
Yes, even after my code change, you can specify the list of variables to be signed, with the default being none (which in my experience is a good default as it is sufficient to protect from most attacks).

lucadealfaro: If you tell me that you prefer one class, I can make that code change.

lucadealfaro:
# Problem: URL() cannot be used outside of request handlers

Aside from the specific comments in the code, I came up with a major question/problem (that was present already in the previous implementation). 

The URL helper can be used _only_ in a controller, not in a declaration. With this I mean that

```
@action("index")
def index():
    u = URL("something")
```

works, but

```
u = URL("something")
@action("index")
def index():
    ...
```

does not.  The problem is that URL relies on the request object to know the application name.

This leads to the problem that one cannot sign a URL outside of a controller.  This is needed for instance in mtable, where one needs to specify the URL for AJAX to the constructor class of mtable.  I am also developing my own version of "grid", which I will be happy to share, and the same problem is present there; I need to do something like: 
```
mygrid = MyGrid(..., callback_url=URL("mygrid", signer=signer))
```

This needs solving, because in py4web, we need to create these classes in the controllers.py code but not in the request handlers, as we cannot register a callback once for each request. 
There are the following solutions to this. 

## Solution 1: make URL application-specific

The most elegant solution would be to define the URL helper as follows: 
```
APPURL = URLFactory("appname")
```

`URLFactory` would take the name of the app as argument, and it would return an object whose `__call__` method acts as the current `URL()` helper.  One would put the above code in `common.py` presumably.  Does this break things elsewhere? 

## Solution 2: enable the URL helper to figure out the app somehow

I have no idea how to do this, but perhaps it can be done. 

## Solution 3: do not sign the application name

Currently, in a request `/app/controller` we sign the entire string `/app/controller`. We could consider signing only `/controller`, so that it would be possible to sign urls also outside of the context of a request.  

From the point of view of security, there is one attack, in which there are `/app1/controller` and `/app2/controller`, and the user is logged _with the same session_ in both app1 and app2.  I do not believe that this is a realistic attack however.  It can already be done with `https://my.host1.com/app/controller` and `https://my.host2.com/app/controller` if one shares sessions.  So I don't think that there is an issue. 

Still, this would lead us to two ways of generating URLs: 

- via the URL helper, when inside a request handler,
- by hand, in a "poor person way", when outside of a request handler. 

This is not super pretty.

lucadealfaro:
Ok, I have successfully used this code, and I think this PR is good.

py4web/utils/url_signer.py
mdipierro: The philosophical sticky point this: do we want one signer per app or one signer per url? I would prefer one per app because we could, for example, create it by default in Auth, but this requires that some parameters be passed not to constructor but to verify(). If we decide we want one signer per action, then there is reason for verify() and/or two classes. Yet this would be a very unusual fixture because we would need one instance of the fixture for each signed action.

lucadealfaro: Let me argue two things: (a) one almost never needs to sign variables, and so in practice only one signer per app is enough, and (b) even if you need to sign variables, at that point, multiple signers is the best.  This means, the current implementation is the correct one. 

This is of course my opinion, but here is my argument. 

(a) signing variables is generally not necessary. 

This is true for two reasons.  First, if one signs a URL u?v=a, then one expects v to be present also in the subsequent use of the URL, and so, one would use as URL u/a , without the variable v.  That is, variables are generally used for optional parameters, not compulsory ones, and the very fact of signing the URL inclusive of the variable shows that the variable in fact is compulsory.  

Second, the main reason for signing URLs is as AJAX endpoints, and it is much easier to give an endpoint of u or u/a, rather than one of u?v=a, because an endpoint without variables (such as u or u/a) is easier to use in Javascript by adding additional variables as in GET calls. 

Indeed, in my own development and in the classes I taught, the old web2py default of including variables in the signature and verification was a constant cause of headaches. 

(b) even if one wants to include variables in the signature, it is much better to then use multiple signers. 

In fact, suppose you have only one signer per app.  Then to sign you would need something like: 

```
signed_url = URL('/something', vars=dict(a=2, b=3), url_signer=url_signer, signed_vars=["a"])
```

and then to verify: 

```
@action.uses(url_signer.verify(signed_vars=["a"]))
```

This is (in my opinion) simply horrible, as there is a list that you must repeat identically in two different places, a big NONO in my mind.  Much better to create a `url_signer_with_a` once and forall, and use that both in the URL constructor, and in the @action.uses method. 

Of course, this is just my opinion.

mdipierro: OK. I am convinced. :-)

PR #120, opened 2020-02-18
Oauth2 updates

crc1021:
This has three changes:
report more specific error from server if it is available
update google settings to not using people api
allow forcing of url http/https scheme for callback uris


mdipierro:
Thanks for this!

PR #124, opened 2020-03-14
Add ability to edit json fields from dashboard

misl6:
This PR adds a minimum implementation for the `json` field in `_dashboard`, and also adds a `json` field usage example in `apps/examples`.

For future reference:
The JSON editor widget in `mtable` is improvable via something like that:
https://github.com/josdejong/jsoneditor

### DIFF
--- a/apps/_dashboard/static/components/mtable.html
+++ b/apps/_dashboard/static/components/mtable.html
@@ -46,7 +46,7 @@
               {{ option.text }}
             </option>
           </select>
-          <textarea class="textarea" v-else-if="field.type=='json'" @input="function(event){item[field.name] = event.target.value}" v-bind:readonly="!editable">{{item[field.name]}}</textarea>
+          <textarea class="textarea" v-else-if="field.type=='json'" @input="function(event){item[field.name] = parse_and_validate_json(event)}" v-bind:readonly="!editable">{{item[field.name]}}</textarea>
           <div class="error" v-if="editable && field.name in errors">{{errors[field.name]}}</div>
         </td>
       </tr>
--- a/apps/_dashboard/static/components/mtable.html
+++ b/apps/_dashboard/static/components/mtable.html
@@ -46,6 +46,7 @@
               {{ option.text }}
             </option>
           </select>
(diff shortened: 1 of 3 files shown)


mdipierro:
I like this but need a little more testing.

misl6:
Just found out a bug to fix before merge.
I'm moving it to `WIP`.

mdipierro:
I am concerned that the JSON received may not be a string of JSON but python object serialized as a sting. Not sure yet.

misl6:
I'm going back on this one.

Analyzing the `pydal` code, I found out that there's something that could be possibly wrong on `json`. (But I could be wrong)

During validation needs a JSON string (it performs checks by parsing the string, then stores the value in the required format based on the adapter):
https://github.com/web2py/pydal/blob/f74a107e00473b3d18ee407f83ab864753770d15/pydal/validators.py#L422

When the user makes a `*.select()` the stored value is parsed and returned as an object: https://github.com/web2py/pydal/blob/202b55d5e9a4018e23950c60ea56e19324dd8fe0/pydal/parsers/base.py#L136

These two operations creates a incosistency between how the data needs to be sent and how is returned and could lead to issues.

Which one should be the correct way to handle it?

Considering that actually the right way to `insert/update` data is the following:

```python
db.person.insert(
        name="Clark Kent",
        job="Journalist",
        resume=[
            {
                "start_date": "2019-12-20",
                "end_date": "2020-01-20",
                "company": "Save The World Srl",
            },
            {
                "start_date": "2020-02-01",
                "end_date": "2020-03-15",
                "company": "SuperHeroes Services Srl",
            },
        ],
    )
```

I'm all-in for changing the validator part.

mdipierro:
Sorry for not being on top of this. I would like to understand better the use case. The validator exists for when input is a string that contains json and it needs to be parsed. If the input is already an object we should not need the validator. If the validator is called anyway (should it?) and the input is not a string, maybe it should just return (value, None) i.e. ignore validation. Do you agree?

misl6:
I super agree.
Will do the fix and rebase ASAP.

misl6:
@mdipierro I did it.
Needs https://github.com/web2py/pydal/pull/622 to be merged. (Appveyor is failing but seems like a deps issue)
I also did a veryvery basic prevalidation that checks the JSON value on the client side.

PR #127, opened 2020-03-20
Fixed errors on ldap plugin + enabled group management

mau86:


mdipierro:
What is the status of this PR. Was it tested? Can you show it woks? I'd love to get this working.

mau86:
OK, I can tell you in more detail what I did and what remains to do:
- First of all, the whole plugin needs to be rewritten from scratch. What I did are more kind of workarounds to make it work
- Connection and authentication work both in "uid" and in "ad" modes (I have OpenLDAP at home and Active Directory at work). At the moment, 'uid' mode was all I had, but I think I can setup a few accounts on OpenLDAP to test more scenarios.
- I've used following settings:
LDAP_SETTINGS = {
    "mode": "ad",
    "server": ldap_host,
    "base_dn": LDAP or AD Base DN (like "ou=Users,dc=example,dc=com",
    "manage_user": True,
    "manage_groups": True,
    "group_dn": Groups Base DN (like "ou=Groups,dc=example,dc=com",
    "group_member_attrib": "member", (this works for AD, probably should be set as default in code)
}

- I've also had to change something in common.py:ì. In order to know about the "groups" object, I have added a parameter when registering LDAPPlugin: instead of `auth.register_plugin(LDAPPlugin(**settings.LDAP_SETTINGS))`, `auth.register_plugin(LDAPPlugin(db=db, groups=groups, **settings.LDAP_SETTINGS))`.

By the way, Massimo, there is a mistyped variable in settings.py: instead of `LDAP_SETTING` it must say `LDAP_SETTINGS`.

With these corrections I've made it to connect and create user and groups both on OpenLDAP and on Active Directory.

mau86:
OK Massimo, I also committed the two things I told about in my previous message (adding two parameters to LDAPPlugin registration and correcting misspelled variable - both in _scaffold app). Sorry for the separate commits, I'm still practicing with VS Code and github...

PR #130, opened 2020-03-21
Fixes vue dashboard reference field

### LINKED ISSUE #129: Latest pydal master breaks dashboard reference fields
Latest `pydal` master breaks `py4web` dashboard (reference fields).

This commit reverts a needed change: https://github.com/web2py/pydal/commit/202b55d5e9a4018e23950c60ea56e19324dd8fe0

@mdipierro was just a cleanup or I missed something when I introduced this one?

misl6:
Fixes #129


mdipierro:
No longer necessary

misl6:
@mdipierro You sure?
Just updated to current `master` on a development server but I can't still access reference fields without that fix on the dashboard  :/

be2881:
I can also confirm that fix #129 is still needed. Without the dashboard dbadmin is not working with reference field. With the fix applied it works.

PR #132, opened 2020-03-28
Update auth.js

carpaIdea:
When auth component is created the form object results empty. No error message is displayed if the user submits an empty login form immediately.
Istantanea_2020-03-28_13-01-50
With the fix the form object will have the fields filled with empty strings and an error message will be displayed to the user.
Istantanea_2020-03-28_13-03-00

### DIFF
--- a/apps/_scaffold/static/components/auth.js
+++ b/apps/_scaffold/static/components/auth.js
@@ -100,6 +100,7 @@
     };
     auth.created = function() {
         var self = this;
+        self.go(self.page, true);
         window.addEventListener('popstate', function (event) {
                 self.go(event.state.page, true);
             }, false);


misl6:
LGTM

carpaIdea:
The correction also fixes the following error that occurs under the same conditions:
```
ERROR:root:Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/py4web/core.py", line 551, in wrapper
    ret = func(*func_args, **func_kwargs)
  File "/usr/lib/python3.8/site-packages/py4web/core.py", line 512, in wrapper
    ret = func(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/py4web/utils/auth.py", line 222, in responder
    return self.action(
  File "/usr/lib/python3.8/site-packages/py4web/utils/auth.py", line 282, in action
    user, error = self.login(**vars)
TypeError: login() missing 2 required positional arguments: 'email' and 'password'
```

PR #136, opened 2020-03-28
Fix ANSI printout on Windows + catch interrupts like Ctrl-C

nicozanf:
see https://stackoverflow.com/questions/36760127/how-to-use-the-new-support-for-ansi-escape-sequences-in-the-windows-10-console
and https://keyboardinterrupt.org/catching-a-keyboardinterrupt-signal/


nicozanf:
I've better to split changes ...

PR #143, opened 2020-04-04
Makes apps folder optional

nicozanf:
This is needed to allow to start the program from the GUI - and so it makes py4web much simpler for new users.

*** WARNING *** : this changes the way to start the program, and we must also change the documentation accordingly!


mdipierro:
I messed up! I merged this and then reverted it. I think we should discuss this on the mailing list.

PR #144, opened 2020-04-04
mtable - vue 2.6.11 warnings

carpaIdea:
py4web is shipped with version 2.2.0 of vue. If latest stable version of vue (2.6.11) is used, there are alarms.
warnings
The proposed fix solves these alarms.

### DIFF
--- a/apps/examples/static/components/mtable.js
+++ b/apps/examples/static/components/mtable.js
@@ -3,10 +3,10 @@
     var mtable = { props: ['url', 'filter', 'order', 'editable', 'create', 'deletable', 'render'], data: null, methods: {}};
     
     mtable.data = function() {        
-        var data = {url: this.url,
+        var data = {mtUrl: this.url,
                     busy: false,
-                    filter: this.filter || '',
-                    order: this.order ||  '',
+                    mtFilter: this.filter || '',
+                    mtOrder: this.order ||  '',
                     errors: {},
                     item: null,
                     message: '',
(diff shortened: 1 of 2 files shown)


apps/examples/static/components/mtable.js
```diff
@@ -3,10 +3,10 @@
     var mtable = { props: ['url', 'filter', 'order', 'editable', 'create', 'deletable', 'render'], data: null, methods: {}};
     
     mtable.data = function() {        
-        var data = {url: this.url,
```
valq7711, line 6: why not just remove url from data? As I see it is not changed by code

apps/examples/static/components/mtable.js
```diff
@@ -3,10 +3,10 @@
...
+        var data = {mtUrl: this.url,
                     busy: false,
-                    filter: this.filter || '',
-                    order: this.order ||  '',
```
valq7711, line 9: `filter` and `order`  are used as initial values, so it is better to remove them from props at all and use `$attrs` like
`filter: this.$attrs.filter || ''`

carpaIdea: $attrs is not compatible with vue 2.2.0 (the version shipped by py4web  currently) because vue introduced it from 2.4.0

https://vuejs.org/v2/api/#vm-attrs

valq7711:
good reason to migrate to latest stable!

carpaIdea:
Of course, but this is a decision for the framework developers. Any user, like me, can customize the client side of an application made with py4web and create their own components using the tools that best suit them.
The intention of my fix is simply to make the mtable code provided by py4web more compatible with the latest versions of vue without breaking the current framework structure.

valq7711:
Yes, but it is client side, that (according web2py philosofy as I understand it) is more just a scaffold than part of the framework

valq7711:
But the scaffold, that is demo of interaction with server side, so why that demo should be legacy?

carpaIdea:
You're absolutely right. My PR is, in fact, on the example of the grid not on a part of the framework. The examples are used by those who want to understand how some features work and experiment as I did. It seemed right to me to propose an adjustment. The developers will then evaluate what to do with it.

mdipierro:
we will upgrade vue!

PR #149, opened 2020-04-06
NOT a PR -- rather, an example of not working

lucadealfaro:
I tried to deploy on Appengine the _minimal app, and I get an error of 

502 Bad Gateway
nginx

It's not clear to me how to solve, and I am making this PR so one can reproduce the problem.
Note that I have moved some dependencies around from requirements.txt to requirements_gae.txt, but that is independent from the problem.


lucadealfaro:
Problem fixed in another PR.

PR #151, opened 2020-04-09
Fix form handling of readable fields

lucadealfaro:
A field should be sent to DAL for writing only if it is writable _and_ readable.  
Otherwise, fields that are writable but not readable will not have their default values set when a record is created. 

Aside from this, the PR adds iterator methods for the session object, so that it becomes possible to write diagnostic code that prints out the code of the session.


mdipierro:
Thank you Luca.

PR #152, opened 2020-04-09
Make apps the default apps_folder

nicozanf:
This is needed to allow to start the program directly from the GUI - and so it makes py4web much simpler for new users.

After PR web2py#143  was reverted, and the discussion on https://groups.google.com/forum/#!msg/py4web/QKSDCVRX53E/9zMboCJ0BAAJ , this should be fine.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -944,7 +944,7 @@ def wsgi(**args):
 def get_args():
     """Handle command line arguments"""
     parser = argparse.ArgumentParser()
-    parser.add_argument("-a", "--apps_folder",
+    parser.add_argument("apps_folder",
         nargs = '?',
         default = (os.path.join(os.getcwd(), 'apps')),
         help="Path to the applications folder (default=apps)",
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -16,6 +16,7 @@ import logging
 import numbers
 import os
 import getpass
+import pathlib
 import platform
 import sys
 import threading
@@ -943,10 +944,14 @@ def wsgi(**args):
 
 def get_args():
     """Handle command line arguments"""
+    
+    #get the real running path, needed from MacOs GUI
+    running_path = pathlib.Path(__file__).resolve().parent.parent
+    
     parser = argparse.ArgumentParser()
     parser.add_argument("apps_folder",
         nargs = '?',
-        default = (os.path.join(os.getcwd(), 'apps')),
+        default = (os.path.join(running_path, 'apps')),
         help="Path to the applications folder (default=apps)",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


lucadealfaro:
I actually like to have to specify explicitly "apps", it makes obvious what the code is doing. 

And I don't like to have as default a subfolder called "apps".  I have not seen much code where, if you don't specify a folder, the default is _a subfolder with a particular name_.  Usually, when you do not specify a folder, the default is _the current directory_.  But this would be cumbersome in py4web given the typical layout of the apps.  

What I mean is that "apps" is not a reserved word or a name typically used in defaults for folders, so I find the behavior of "python py4web.py" --> "I really mean to run it on the apps folder in the current directory" to be odd, rather uncommon among python things. 

I guess I just don't like reliance on naming conventions, as they tend to make the control flow obscure.  

This is obviously purely my opinion (I use py4web in teaching).

mdipierro:
I tend to agree with Luca on this one. Is there any reason we need a default value? Can this be handled by the windows/osx binaries?

nicozanf:
The problem is that you cannot specify any option when you start a program from the GUI. So, if there are mandatory options you will never be able to use a program from the GUI ;-(

But maybe I've found an acceptable compromise: leave it optional but if not provided ask for its value (with 'apps' as default to be confirmed). In this way running from GUI will work AND the apps_folder is clearly stated and changeable at startup.

Does it sound good for you?

lucadealfaro:
The problem with a GUI app that asks where the apps are is that some apps come with the distribution; first of all, _dashboard.   So what can one do, put _dashboard manually in another apps folder?  This would defeat the simplicity of the GUI app.  Or read from both the internal apps folder and the specified one?  But then you are back at square zero. 

I guess, it would have been cleaner if py4web came with no apps, and then one had a separate repository of apps, of which _dashboard, _examples, and others would be instances.  The GUI app could ask you for a folder, and it could ask you whether you want to initialize it by copying in it the _dashboard app. 

So I don't know what to say. 

On the one hand, you can make the argument optional.  I don't like the --apps_folder option, but I would be ok if 

```
python py4web-start.py apps
``` 
uses the specified folder, and 
```
python py4web-start.py 
```

uses the one where py4web-start.py is. 

All I can tell you is that when I teach, I don't use the GUI app for two reasons.  One is that the py4web code is still in flux and I want students to be able to pull from the git repo, but this is true I guess this year only.  The other reason is that students are typically confused (they were confused with web2py) about where the apps were.  When I distributed a homework by telling them "clone this git repo in your apps folder" they were confused how to navigate with the command line to the web2py (then) folder in /Applications/.... in order to type the git commands.

mdipierro:
We decided to go with this instead #153

PR #154, opened 2020-04-12
Added signature to forms; used hmac

lucadealfaro:
This PR should likely be well reviewed / discussed as it is a bit crucial. 

The PR adds signatures to forms.  To ask for a form to be signed, one adds a csrf_session=session argument to it.  This inserts a _single_ signing key into the session, called _form_key; the same key is used for all forms.  This prevents blowing up the session using one key per form, as was the case in web2py, where the forever-growing session size would eventually blow up sessions stored into cookies. 

Once this is done, a form has: 

* A _formname, used in case more than one form is included in the page. 
* A _formkey, computed by signing some salt.  The salt, and its signature, form the _formkey. 

The _formkey and _formname are verified. 

I also modified the behavior, and if the verification fails, the form now behaves like a GET.  This ensures that, if there are multiple forms on a page, the form whose verification fails (because the post is for the other form) is again re-presented with the data, as if one had requested it fresh. 

I also made both form signatures and URL signatures use hmac. 

**Problem:** If a hmac_session is specified, then we must have @action.uses(session).  I cannot make this form be a Fixture, because this type of forms is created inside a controller, and so, it is not available at the time the @action occurs.  If one forgets the (session) argument, then the key is not saved in the session, and verification typically fails. 

The forms I created for Vue do not have this problem, but I digress.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -175,20 +175,22 @@ class Form(object):
     - dbio: set to False to prevent any DB writes
     - keep_values: if set to true, it remembers the values of the previously submitted form
     - form_name: the optional name of this form
+    - csrf_session: if None, no csrf token is added.  If a session, then a CSRF token is added and verified.
     """
 
     def __init__(
-        self,
-        table,
-        record=None,
-        readonly=False,
-        deletable=True,
-        formstyle=FormStyleDefault,
-        dbio=True,
-        keep_values=False,
-        form_name=False,
-        hidden=None,
-        validation=None,
+            self,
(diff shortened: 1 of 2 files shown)


lucadealfaro:
The reason why it may be desirable to sign a form is that sameorigin=Strict is not necessarily sufficient against CSRF.  For instance, if you have a webhosting.com domain that is not included in the [Mozilla public suffix list](https://publicsuffix.org/learn/), then site1.webhosting.com can do a CSRF to site2.webhosting.com .

PR #166, opened 2020-04-22
fix to form initialization when form is created as list of Field

lucadealfaro:
When a form is defined by a list of Field , rather than an actual table, the initialization of the form values via a record= was broken.  This fixes it.  
One residual problem is that I no longer use: 

```
table[name].formatter(self.record[name])
```
because I don't have a real table; so I just do:
```
field.name: self.record.get(field.name)
```

I hope this works, otherwise feel free to improve.


lucadealfaro:
Does this break the CI??

PR #176, opened 2020-05-03
jwt signatures for URLs and forms

lucadealfaro:
This PR adds jwt signatures for URLs and forms, which:

- enables to specify an expiry time
- enables to add user information (not leaked to the token) or other information to the signing key. 

Overall, this is cleaner than the hmac implementation, and more extensible. 

The one point that can be improved is that so far, if a form fails verification, there is no error displayed to the user.  Since the ".accepted" attribute of the form is set to False, the user is simply displayed a new form, as if the user did a GET on the page.  If a form is expired, or fails verification, should a better mechanism be implemented to notify the user that the form submission was rejected?

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -1,6 +1,8 @@
+import json
+import jwt
+import time
 import uuid
-import hashlib, hmac
-from py4web import DAL, request
+from py4web import request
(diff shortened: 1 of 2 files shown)


mdipierro:
Thank you Luca!

PR #183, opened 2020-05-07
Added functionality to clear tickets from the dashboard

pdralston:
This functionality adds a button to the _dashboard's "Recent Tickets" and allows you to clear the tickets.


mdipierro:
It would be nice it there were also buttons that given a traceback would clear only matching tickets. :-)

pdralston:
@mdipierro I will look into that.

PR #193, opened 2020-05-14
Allowed for more granular syntax highlighting

agavgavi:
Highlight.js will by default try to figure out which file type it is rendering, but sometimes that will have bugs. To make it more granular and specific, a function was added that will set the class of the `<code>` tag to the specific language we need, making sure it will enable the correct highlighting. Because of that, if users want to use gitlog on more file types, to they can just add another case to the function that will correspond to their specific application.

In addition, allowed for more text to be visible in the commit message since it can be difficult to recognize which commit is which with only the first 20 characters.

### DIFF
--- a/apps/_dashboard/diff2kryten.py
+++ b/apps/_dashboard/diff2kryten.py
@@ -21,8 +21,9 @@ html {background-color: white; font-family: helvetica}
 
 script = """
 hljs.initHighlightingOnLoad();
-$('.line:not(.line-new)').each(function(){$(this).text($(this).attr('data-content'));});
+$('.line:not(.line-new)').each(function(){$(this).text($(this).attr('data-content') + "\\n")});
 $('.line-new').hide();
+$('.hide-newline').hide();
 $('.diff').hide();
 $('.file .filename').click(function(){$(this).closest('.file').find('.diff').slideToggle();});
 var block = 0;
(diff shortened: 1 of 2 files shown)


apps/_dashboard/diff2kryten.py
```diff
@@ -77,6 +77,20 @@ def escape(txt):
         .replace('"', "&quot;")
     )
 
+# specify which language we want to render as for highlight.js
+# based on the name of the file
+def getFileType(name):
+
+    if ".py" in name:
+        return 'class="language-kotlin"'
```
mdipierro, line 85: good idea but this should probably return Python

apps/_dashboard/diff2kryten.py
```diff
@@ -77,6 +77,20 @@ def escape(txt):
...
+    if ".js" in name:
+        return 'class="language-coffeescript"'
```
mdipierro, line 87: should return javascript

apps/_dashboard/diff2kryten.py
```diff
@@ -77,6 +77,20 @@ def escape(txt):
         .replace('"', "&quot;")
     )
 
+# specify which language we want to render as for highlight.js
+# based on the name of the file
+def getFileType(name):
+
+    if ".py" in name:
```
mdipierro, line 84: Should check for name.lower().endswith('.py')
and so the other ones.

mdipierro:
Thanks for doing this. Would be great if you can make the minor corrections which I suggested.

agavgavi:
I'm sorry, I'm still new and learning how to use pull requests, but I'm having a hard time locating the corrections, I've tried looking on the commit but I don't see and comments or anything that would signify a correction? I'm sure I'm just missing the button or tab that I need to be looking at though.

mdipierro:
I think I did it wrong. Can you try now?

agavgavi:
I do see it now, I will make these changes and push, the reasoning behind coffeescript and kotlin was due to the syntax highlight was having a hard time recognizing the languages, so if you use python or javascript it would not highlight anything. I found these to be the closest to proper highlighting and thought it best to use that until I had slightly more time to research exactly why this issue occurs. I have a suspicion it has to do with the theme I chose but I need to look more in depth first.

mdipierro:
Interesting. If using python and js does not work for you leave those as they are.

agavgavi:
I've added the names parsing and I will be researching this weekend more indepth with highlight.js

agavgavi:
Proper syntax highlighting should work now, after spending many hours in different JSFiddle testing grounds, I came to the realization that the reason proper syntax highlighting did not work with the correct languages was because of the use of divs for classes `line line-old`/`line line-new`/`line`. 

This seems to be because of how Highlight.js (and other syntax highlighters work, I tried a bunch) likes to read their code. When I swapped the types of these to ``span`` it worked as intended. The only issue after that was adding the line breaks that were lost when the swap from divs occurred. 

By default all code would appear on one line, and adding a "\n" after each line made large empty blocks where the new lines would eventually go. and personally I found the best option (and the one that in use looks the most similar to the original application) was to show and hide an additional span for the line-new that would create a space for the line to go before it begins to be written. 

Another option I tried that ended up not working was adding a newline character after each time we move the data-content field to the text field, the issue from this was that with the "typing" mechanism that exists, it would begin displaying the line on the same line as the next actual line, moving the characters instead of creating a new line to write on.

It is because of this reason that I feel that the conditional span may be the best idea on how to approach this.

This does allow for `language-python` and `language-javascript` to work perfectly in setting classes, so that should mean that future language expansion should be simple now that this base has been developed.

Please let me know if there is anything you think should be implemented differently, but hopefully this will end the saga of syntax highlighting that has been occurring the past few days.

Issue #108, opened 2020-02-09
Each worker has own session?


valq7711:
https://github.com/web2py/py4web/blob/830c76a2222d0a62a37bea8a5bdbfd9bbb5ddda4/py4web/core.py#L331

In case of usage multiple workers, each worker will have its own secret  i.e. own session! (If `secret`-argument is not provided).
I got this issue on gunicorn with gevent workers. 
Could we store `uuid` key in `os.environ` to share secret between workers? 

UPDATE:
As variant, we can store in os.environ only the flag like `secret_is_generated`, if it is not set then uuid-secret should be generated and stored in some file, otherwise, the `secret` should be read from that file.

mdipierro:
This is a serious bug. Will address tomorrow.

PR #194, opened 2020-05-16
Add --watch option

chbdetta:
It is convenient for developer if the server can automatically reload an app once some python files inside that app are changed.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -963,9 +963,9 @@ def watch_folder_event_loop(apps_folder):
 async def watch_folder(apps_folder):
     click.echo('watching python file changes in: %s' % apps_folder)
     async for changes in awatch(os.path.join(apps_folder)):
-        for app in set(map(lambda p: p.relative_to(apps_folder).parts[0],
-                        filter(lambda p: p.suffix == '.py',
-                                map(lambda pair: pathlib.Path(pair[1]), changes)))):
+        for app in set([p.relative_to(apps_folder).parts[0]
+                for p in [pathlib.Path(pair[1]) for pair in changes]
+                if p.suffix == '.py']):
             Reloader.import_app(app)
 
 def start_server(args):
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -31,6 +31,8 @@ import types
 import urllib.parse
 import uuid
 import zipfile
+import asyncio
+from watchgod import awatch
 
 REX_APPJSON = re.compile(r"(^|\s|,)application/json(,|\s|$)")
 
@@ -953,13 +955,32 @@ def error404(error):
 # Web Server and Reload Logic: Operations
 #########################################################################################
 
+def watch_folder_event_loop(apps_folder):
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    loop.run_until_complete(watch_folder(apps_folder))
+
+async def watch_folder(apps_folder):
+    click.echo('watching python file changes in: %s' % apps_folder)
+    async for changes in awatch(os.path.join(apps_folder)):
+        for app in set(map(lambda p: p.relative_to(apps_folder).parts[0],
+                        filter(lambda p: p.suffix == '.py',
+                                map(lambda pair: pathlib.Path(pair[1]), changes)))):
+            Reloader.import_app(app)
 
 def start_server(args):
-    host, port = args['host'], int(args['port'])
+    host, port, apps_folder, watch = args['host'], int(args['port']), args['apps_folder'], args['watch']
+
     if platform.system().lower() == "windows":
+        if watch:
+            # default wsgi server block the main thread so we open a new thread for the file watcher
+            threading.Thread(target=watch_folder_event_loop, args=(apps_folder,)).start()
         # Tornado fail on windows
         bottle.run(host=host, port=int(port), reloader=False)
     elif args['number_workers'] < 1:
+        if watch:
+            # tornado delegate to asyncio so we add a future into the event loop
+            asyncio.ensure_future(watch_folder(apps_folder))
         bottle.run(server="tornado", host=host, port=port, reloader=False)
     else:
         if not gunicorn:
@@ -967,6 +988,9 @@ def start_server(args):
         elif not gevent:
             logging.error("gevent not installed")
         else:
+            if watch:
+                click.echo('--watch option has no effect in multi-process environment \n')
+
             sys.argv[:] = sys.argv[:1]  # else break gunicorn
             bottle.run(
                 server="gunicorn",
@@ -1059,7 +1083,6 @@ def fix_ansi_on_windows():
         from ctypes import windll
         windll.kernel32.SetConsoleMode(windll.kernel32.GetStdHandle(-11), 7)
 
-
 def keyboardInterruptHandler(signal, frame):
     """Catch interrupts like Ctrl-C"""
     click.echo("KeyboardInterrupt (ID: {}) has been caught. Cleaning up...".format(signal))
@@ -1111,7 +1134,6 @@ def set_password(password, password_file):
     with open(password_file, "w") as fp:
         fp.write(str(pydal.validators.CRYPT()(password)[0]))
 
-
 @cli.command()
 @click.argument('apps_folder', default='apps')
 @click.option('-Y', '--yes', is_flag=True, default=False, help='No prompt, assume yes to questions')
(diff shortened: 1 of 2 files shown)


valq7711:
Hi chbdetta!
I also feel lack of this option, but as I understand it will invoke reloading on each ctrl-s?

chbdetta:
@valq7711 Yes, It will. And I think that is acceptable because it only reloads one app and the cost is not huge. Also --watch option should only be used in development, so no overhead after the app is deployed. Also I think triggering a reload on each ctrl-s feels predictable for users.

valq7711:
Maybe it is better to just mark app as 'need to be reloaded' and reload on request if required?

valq7711:
I mean we can check for out_of_date apps on each request

chbdetta:
By 'reloading on request' do you mean a network request? But that only works with codes that run after a request. In situations like adding a print function in a model.py file, I think it's predictable to see the logging immediately instead of triggering the change using a page reload to send a request.

The purpose of '--watch' is for developers to see changes in their code without manually restarting or reloading the apps. If they need to send a request to the server to see the changes, that kind of defeat the purpose itself.

valq7711:
> By 'reloading on request' do you mean a network request? 
yes, on each request that py4web receives

> By 'reloading on request' do you mean a network request? But that only works with codes that run after a request. 
It is not so, there is `bottle.add_hook('before_request', hook)`  - such hooks are processed before routes mounting, I mean like the following: 
```python
#core.py
def install_reload_hook():
    def hook(*a, **k):
        '''
        check for out of date apps, that need to be reloaded 
        out_of_date_apps - list of app names
        also we can use request.path to detect which app is requested at the moment
        and reload only it, but not all from out_of_date
        '''
        for a in out_of_date_apps:
            Reloader.import_app(a)
            # mark somehow that `a`-app is up to date now 
            
    bottle.default_app().add_hook('before_request', hook)    
        
    
# if --watch in CL-args  install_reload_hook()

**Comment by chbdetta** [comment]
Yeah. If I understand correctly, it would not reload any app until the developer send a request to the server?
If I add a `print` function in the model.py file, I would only see the output when I trigger some requests to the server.

**Comment by chbdetta** [comment]
Also, the implementation you gave would reload apps that are marked to be out_of_date before a request? That means if app `a`, `b`, `c` are marked to be reloaded and I send a request to app `a`, it would reload all three apps instead of just `a`?

**Comment by chbdetta** [comment]
I just don't really understand what the advantage is to delay the reload until a request is sent and is it worth making the `--watch` behavior not so automatic in some situations?

**Comment by valq7711** [comment]
>That means if app a, b, c are marked to be reloaded and I send a request to app a, it would reload all three apps instead of just a?

as I mentioned in the hook-comment we can detect requested app using request.path and reload only it.

>I just don't really understand what the advantage is to delay the reload until a request is sent and is it worth making the --watch behavior not so automatic in some situations?

in some situations:
-  application may be quite large
-  sometimes developers edit several files at once - while fixing something here, suddenly remembered that something needs to be fixed  there - in this mode, saved files may just have invalid syntax or unexpected/unsafe behavior

in any case, developers should have a choice

**Comment by valq7711** [comment]
also, reloading causes small memory leaks - I have some thoughts need to be tested

**Comment by chbdetta** [comment]
> application may be quite large
I agree that the reloading time may be long if the application is large. But delaying the reloading until a request means extra waiting time for an unresponsive server. Reloading immediately starts when file is saved can reduce such wait. I think we could add debounce to prevent too frequent reloading of an app.

> sometimes developers edit several files at once - while fixing something here, suddenly remembered that something needs to be fixed there - in this mode, saved files may just have invalid syntax or unexpected/unsafe behavior

This is exactly why reloading immediately after saving is good. Developers can see feedback right after they modifying the codes instead of until triggering a request to the server. Switching between context (the code editor and the browser) is not a smooth experience, which is what this PR aim to solve.

> also, reloading causes small memory leaks - I have some thoughts need to be tested

I believe this should be fixed in the Reloader instead of delaying the feedback of code changes.

**Comment by valq7711** [comment]
>This is exactly why reloading immediately after saving is good. Developers can see feedback right after they modifying the codes instead of until triggering a request to the server

;)) I mean a slightly different case - the intentional saving of unfinished files (with invalid syntax or  wrong behavior) - you just suddenly have to go somewhere

**Comment by chbdetta** [comment]
> ;)) I mean a slightly different case - the intentional saving of unfinished files (with invalid syntax or wrong behavior) - you just suddenly have to go somewhere

In such case having an error message pop up every time you save reminds you the unfinished invalid syntax is still there waiting to be fixed before you go test in the browser :D

**Comment by valq7711** [comment]
:D, ... so can we add something like `--watch=lazy`?

**Comment by chbdetta** [comment]
Maybe make it a separate PR? I think this PR should solely focus on the basic watching ability

**Comment by valq7711** [comment]
>Maybe make it a separate PR? I think this PR should solely focus on the basic watching ability
OK, sure

File: py4web/core.py [hunk 1]
```diff
@@ -953,20 +955,42 @@ def error404(error):
...
+    asyncio.set_event_loop(loop)
+    loop.run_until_complete(watch_folder(apps_folder))
+
+async def watch_folder(apps_folder):
+    click.echo('watching python file changes in: %s' % apps_folder)
+    async for changes in awatch(os.path.join(apps_folder)):
+        for app in set(map(lambda p: p.relative_to(apps_folder).parts[0],
+                        filter(lambda p: p.suffix == '.py',
+                                map(lambda pair: pathlib.Path(pair[1]), changes)))):
+            Reloader.import_app(app)
```
valq7711, line 970: just one question;): 
It looks very javastriptish  - maybe it is related to `async` somehow?
it would be more pythonic something like:
```python
for app in set([p.relative_to(apps_folder).parts[0]  
        for p in [pathlib.Path(pair[1]) for pair in changes] 
        if p.suffix == '.py'])
    Reloader.import_app(app)
```

chbdetta: Yeah I am more familiar with iterator based style than list comprehension :), unless this is a hard requirement to be pythonic, I think it works well?
Also, wouldn't this create two intermediate list?

py4web/core.py
```diff
@@ -953,20 +955,42 @@ def error404(error):
...
+            # tornado delegate to asyncio so we add a future into the event loop
+            asyncio.ensure_future(watch_folder(apps_folder))
         bottle.run(server="tornado", host=host, port=port, reloader=False)
     else:
         if not gunicorn:
             logging.error("gunicorn not installed")
         elif not gevent:
             logging.error("gevent not installed")
         else:
+            if watch:
```
chbdetta, line 991: Actually, I want to point out here I didn't add the --watch ability when args['number_workers'] > 0 because I think users only use the number_workers option to deal with high server load in production, which is contracting the usage context of `--watch`

py4web/core.py
chbdetta: I think the syntax is clearer using list comprehension, I made a commit on that. Thank you!

py4web/core.py
```diff
@@ -953,20 +955,42 @@ def error404(error):
...
+                click.echo('--watch option has no effect in multi-process environment \n')
+
```
chbdetta, line 993: I don't really want to put the watch logic in start_server function. But the watcher initialization is affected by which server is used and I don't like to re-implement the same condition checking elsewhere either. Any advice on this?

py4web/core.py
```diff
@@ -953,20 +955,42 @@ def error404(error):
...
```
chbdetta, line 993: Actually, I want to point out here I didn't add the --watch ability when args['number_workers'] > 0 because I think users only use the number_workers option to deal with high server load in production, which is contracting the usage context of --watch.

Please correct me if this is not the case.

py4web/core.py
valq7711: maybe place it in the separate module and do some like:
```python
def start_server(args):
    ...
   if watch and args['number_workers'] < 1: 
      import Watcher from whatever
      watcher = Watcher(Reloader, apps_folder, platform.system().lower())
  
```

py4web/core.py
valq7711: I agree with you, this is too much in the case of several workers
... but `shit happens`:D:  https://github.com/web2py/py4web/issues/108

mdipierro: I agree.

chbdetta:
@lucadealfaro I am happy to do that, but the current README seems outdated compared to the CLI commands.

PR #198, opened 2020-05-18
Added the ability to create new files from dashboard app

ch-rigu:
Hi, 
I've write the function that creates new files from the admin dashboard :), adapt it if necessary.
btw, you mind if I create a option to select a different appearance (future.css) ?. The actual colors are hard to use during the day... would be nice to have different appearances to the admin dashboard.

Cheers.
Chris.

### DIFF
--- a/apps/_dashboard/__init__.py
+++ b/apps/_dashboard/__init__.py
@@ -17,6 +17,7 @@ from pydal.validators import CRYPT
 from yatl.helpers import BEAUTIFY
 from .utils import *
 from .diff2kryten import diff2kryten
+from pathlib import Path
 
 MODE = os.environ.get("PY4WEB_DASHBOARD_MODE", "none")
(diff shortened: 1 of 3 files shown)


mdipierro:
Thanks. Good job. Made some minor changes. Please check I did not break it. ;-)

ch-rigu:
Hi Massimo, it works but is trying to write in any place in the file system, if you pass an argument starting with "/" like: /blabla/file.html, it will try to create the folder in the system root directory.
Do you want me to fix this or you will do it ?

Cheers.
Chris.

mdipierro:
My bad! can you fix it? something like this:

fullpath = os.path.join(path, filename)
if not paths.startswith(path + os.sep): error

ch-rigu:
> My bad! can you fix it? something like this: fullpath = os.path.join(path, filename) if not paths.startswith(path + os.sep): error

Sure! I'll do it.

PR #199, opened 2020-05-19
Luca better form

lucadealfaro:
Big improvements to vueform js forms: 
- Can now understand datetime (not 'date', not 'time', so far, just 'datetime') translating between UTC on the back end and localtime in the browser
- Supports IS_IN_SET with dropdowns
- supports writable=False and readonly=True for the form. 

I use luxon as js package to do localtime to UTC conversions and time formatting. 

I need to add 'date' and 'time' alone, even though I hardly ever use them, because a date is not the same date in all timezones; only a datetime has a universal meaning.  So I am not sure it is high priority (it is not for me).


mdipierro:
Please let me know if this is ready for merge

lucadealfaro:
I wanted your opinion first on which way to define components, see the
email I sent you.

mdipierro:
I am ready to merge this if you confirm it is ready.

PR #203, opened 2020-05-20
minor fix in create new file

ch-rigu:
now is being checked the path


ch-rigu:
my mistake  using git... heheh, will do a new pr

Issue #167, opened 2020-04-25
db.table.with_alias - problem (it is not thread.local)


valq7711:
I got the unexpected (for me at least) behaviour of `with_alias` feature which is in fact  means `set_alias` or more precisely `set_alias('alias', overwrite_existing = True)`   - I mean that it mutates `db`, so  if you have predefined  table `foo` then `db.bar.with_alias('foo')`  kills/replace  `db.foo`
Regarding py4web, this problem is even worse becasue `with_alias` changes `db` globally,  so all defined aliases in any requests will persisted between requests/controllers and live  until server restart!

valq7711:
fixed in pyDAL

PR #207, opened 2020-05-20
fix issue #167

valq7711:
Clear tables aliases before each request
**requires the last version of pyDAL**  (with [this PR](https://github.com/web2py/pydal/pull/630) included)

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -282,6 +282,7 @@ class DAL(pydal.DAL, Fixture):
 
 
 # make sure some variables in pydal are thread safe
+setattr(pydal.DAL, "_aliased_tables", threadsafevariable.ThreadSafeVariable())
 for _ in ["readable", "writable", "default", "update", "requires"]:
     setattr(pydal.DAL.Field, _, threadsafevariable.ThreadSafeVariable())
 
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -272,6 +272,7 @@ class Translator(pluralize.Translator, Fixture):
 class DAL(pydal.DAL, Fixture):
     def on_request(self):
         threadsafevariable.ThreadSafeVariable.restore(ICECUBE)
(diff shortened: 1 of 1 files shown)


py4web/core.py
```diff
@@ -272,6 +272,7 @@ def on_success(self):
 class DAL(pydal.DAL, Fixture):
     def on_request(self):
         threadsafevariable.ThreadSafeVariable.restore(ICECUBE)
+        self._aliased_tables.clear() # should we do tha same on_error/on_success?
```
mdipierro, line 275: Damn if we need to clear it, _aliased_tables is not thread safe. Need to check before merging this.

valq7711: Oh yeah, my bad, is it fixed now? (with [below commit](https://github.com/web2py/py4web/pull/207/commits/d82afd1df460cf0650bda88c46af343e633abfbb))

mdipierro: testing a fix

mdipierro:
I think this is now fixed in pydal. Can you please confirm?

valq7711:
Yes, now it works just fine! Thanks a lot!

jpsteil:
I'm still seeing an issue.  I have the following form:

```
form = Form([Field('search', length=50, default=search_filter),
         Field('active', 'boolean', label='Active Only', default=active_filter),
         Field('test_date', 'date', requires=IS_DATE('%m/%d/%Y'))],
        keep_values=True, formstyle=FormStyleBulma)
```
In the screenshot you can see it is giving an error even though it is displaying it correctly.  From my reading this is because the HTML5 date control displays your date in your current locale format but always deals with it in the form field as YYYY-MM-DD format.  The pydal validator wants to read it in the format that you specify on the IS_DATE validator.  I hope I'm making sense.

-Jim

Screenshot from 2020-05-26 16-42-48

Issue #202, opened 2020-05-19
Watchgod issues


pdralston:
With the latest pull, watchgod is enabled. I have observed 2 issues with it so far.
1. When running the server using anaconda prompt, pressing Ctrl+C fails to interrupt the server as expected. Pressing the command twice seems to shut down the server as expected. This behavior is not observed when using WSL or PyCharm
2. When testing watchgod functionality, I changed a minor value in a python value, the watcher saw the change and reported that it reloaded the application in question. Upon hard refreshing I get a 500 error. Restarting the server manually allows the application to load in the browser as expected.

pdralston:
![pycharm_setup_9](https://user-images.githubusercontent.com/25166909/82371015-56fdae80-99ce-11ea-9e2a-4d9489363d64.png)

chbdetta:
> When running the server using anaconda prompt, pressing Ctrl+C fails to interrupt the server as expected. Pressing the command twice seems to shut down the server as expected. This behavior is not observed when using WSL or PyCharm

I think this happens only when WSGI server is used and it probably relates to the threads not exiting. 

> When testing watchgod functionality, I changed a minor value in a python value, the watcher saw the change and reported that it reloaded the application in question. Upon hard refreshing I get a 500 error. Restarting the server manually allows the application to load in the browser as expected.

the --watch option basically just uses the Reloader to reload apps after files change. So this bug should be able to be reproduced using the dashboard reload button.

pdralston:
Commented on the commit. Shutting down the server when using anaconda still requires two keyboard interrupts.

pdralston:
@chbdetta On reloading the apps, it is not able to be duplicated from the _dashboard because the dashboard is reloading all apps, not just the singular that was changed. It appears that there is some conflict that is raised by only reloading a singular app.

pdralston:
![watcher_issue](https://user-images.githubusercontent.com/25166909/82845467-967a3e00-9e99-11ea-966a-076177f1d9cf.gif)

pdralston:
@chbdetta see above

mdipierro:
Can you cut and paste the error in the console? Cannot see from the video.

pdralston:
```
ERROR:root:Traceback (most recent call last):
  File "D:\My Documents\School Stuff\CSE183\Projects\py4web\py4web\core.py", line 568, in wrapper
    [obj.on_request() for obj in fixtures]
[...]

PR #208, opened 2020-05-20
add lazy watching

valq7711:
usage:
`py4web run --watch=off` - no watching (default)
`py4web run --watch=sync` - watch in sync-mode (app is reloaded immediately) 
`py4web run --watch=lazy` - watch in lazy-mode (app is reloaded on first incoming request, BUT before request processing)

### DIFF
--- a/apps/_documentation/static/chapters/en/chapter-01.mm
+++ b/apps/_documentation/static/chapters/en/chapter-01.mm
@@ -111,6 +111,57 @@ py4webt run -p password.txt apps
 ### Command line options
 
 py4web provides multiple command line options which can be listed with ``--help``.
+``
+``
+py4web --help
+Usage: py4web.py [OPTIONS] COMMAND [ARGS]...
+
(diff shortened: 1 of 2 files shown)


mdipierro:
Can you please add the usage to chapter 1 in the _documentation?

agavgavi:
Does the implementation of lazy and sync have the same bug as seen on Issue #202 where the server will not properly closed without multiple CTRL+C on things such as the Anaconda prompt? I haven't had a chance to understand watchgod so I don't know what the bug may be, but just wanted to see if that issue was addressed here.

valq7711:
@agavgavi I believe my watchgod-experience is no deeper than yours;). This PR does not change implementation of watching itself, so I think If your bug is related to watchgod, it will still be:(

valq7711:
> Can you please add the usage to chapter 1 in the _documentation?

@mdipierro  OK, sure

PR #213, opened 2020-05-24
fix watcher thread

valq7711:
`daemon=True` means unconditional thread termination without waiting for daemon to complete its work, so does it actually solve the problem? @pdralston

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -997,7 +997,7 @@ def watch(apps_folder, server = 'default', mode = 'sync'):
     
     if server == 'default':
         # default wsgi server block the main thread so we open a new thread for the file watcher
-        threading.Thread(target=watch_folder_event_loop, args=(apps_folder,)).start()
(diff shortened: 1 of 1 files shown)


agavgavi:
Talking with people who had this issue, the new PR caused a new problem that doesn't let py4web even run when using the Anaconda prompt. However on WSL it seems to work just fine from my experience. (note this isn't my screenshot, it is of someone who uses Anaconda) . So no I do not think the previous change solves the problem.
image

valq7711:
the problem with py4web launch happens because of conflict in this [PR](https://github.com/web2py/py4web/pull/210) - please, try with this (my) fix

agavgavi:
Confirmed that your fix addresses that issue. On the latest py4web with your edits it apparently works.

mdipierro:
Thanks for the great collaboration here!

pdralston:
Apologies on the late response btw, I had to take some mental health days, away from the comp.

PR #214, opened 2020-05-27
fix modules clear up when reloading apps fails

chbdetta:
`Reloader.import_app()` doesn't clear the submodules of an app in `sys.modules` when an error occurs during importing, leaving stale submodules in the cache and not able to be properly reloaded.


mdipierro:
good catch!

PR #215, opened 2020-05-27
Cleaned up the branch button on chrome and other attributes for Gitlog

agavgavi:
Using the a tag on chrome even with -webkit-appearance as an attribute would mis-represent the button, so it was better to replace it with a form submit button.

In addition, the line wrapping for too large commit messages and branch names lead to the Checkout and Show buttons extending past the width of the selector panel, so that size was shrunk.

__See below comment for update about dangerous branches after this post__
Finally, there is currently an issue with how the requests to swapping the branch work that if a branch has a / in it (ex: agavgavi-branches/branch-1) the URL will read the slash as a path delimiter. To address this issue for now, all branches with these types of path are not included in the list and users will have to git checkout themselves. However, I am currently working on developing a way to make it such that the branch will be readable in a reasonable manner. However, I personally found that this method would be a good solution for now because the users would not get 404 errors when using the application.

### DIFF
--- a/apps/_dashboard/__init__.py
+++ b/apps/_dashboard/__init__.py
@@ -60,12 +60,6 @@ def get_branches(project):
     output = run("git branch", project)
     branches = {"current" : "", "other" : []}
     for line in output.split("\n"):
-
-        # until able to figure out how to send branches like py4web/test-branch
-        # those branches must be checked out manually through the users git tool
-        # of choice
-        if "/" in line:
-            continue
         if line.startswith("* "):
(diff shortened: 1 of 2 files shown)


agavgavi:
Strike the last part about requests with /. By converting the form to a full POST request form, the branch can be sent as a parameter through form data that can be fished out by the controller in __init__.py This allows any branch title that bottle can handle in terms of URI conversion. In addition it simplifies some code and makes other areas a little cleaner.

PR #216, opened 2020-05-28
Added help strings to Click interface

nicozanf:
+ nogui option

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -1154,8 +1154,9 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '-help', '--help'])
 def cli():
     pass
 
-@cli.command(help='Show versions and exit')
+@cli.command()
 def version():
+    """Show versions and exit"""
     from . import __version__
     click.echo('\npython -> ' + sys.version)
     click.echo('system -> ' + platform.platform())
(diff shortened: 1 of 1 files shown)


valq7711:
@nicozanf it seems that you use out of date core.py as base

nicozanf:
> @nicozanf it seems that you use out of date core.py as base

right: I've restored the updated watch option, thanks @valq7711  ;-)

py4web/core.py
```diff
@@ -1138,65 +1141,87 @@ def keyboardInterruptHandler(signal, frame):
...
 
 @cli.command()
-@click.argument('apps_folder', default='apps')
-@click.option('-Y', '--yes', is_flag=True, default=False, help='No prompt, assume yes to questions')
-@click.option('-H', '--host', default='127.0.0.1', help='Host name')
-@click.option('-P', '--port', default=8000, type=int, help='Port number')
-@click.option('-p', '--password_file', default='password.txt', help='File for the encrypted password')
-@click.option('-w', '--number_workers', default=0, type=int, help='Number of workers')
-@click.option('-d', '--dashboard_mode',  default='full', help='Dashboard mode: demo, readonly, full (default), none')
+@click.argument('apps_folder', default='apps', type=click.Path(exists=True))
```
valq7711, line 1212: @nicozanf you're welcome!
But why do you require apps-folder to exist? (for run-command at least)
If there is no apps-folder, it is created during `intstall_args`
I think this is causing the test to fail     
https://github.com/web2py/py4web/blob/e92dc89c841e958c822ab1227476b06b8e9bb537/py4web/core.py#L1070-L1085

nicozanf: I misses it, thank you again @valq7711 !

py4web/core.py
```diff
@@ -1138,65 +1141,87 @@ def keyboardInterruptHandler(signal, frame):
     click.echo("KeyboardInterrupt (ID: {}) has been caught. Cleaning up...".format(signal))
     sys.exit(0)
 
+if hasattr(sys, 'frozen'):
+    # PyInstaller
+    py4web_cmd = 'py4web-start'
```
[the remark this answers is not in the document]
nicozanf: OK, it's better to avoid this - it should not depend on the packaging method.

PR #221, opened 2020-06-20
Better components and documentation

lucadealfaro:
This implements better components, with an url() methods so that they can be instantiated from vue.  I have also added quite a bit of documentation to the components.  Also, the vue grid has a search bar, and the table fields are sortable.


apps/examples/components/README.md
```diff
@@ -32,28 +30,133 @@ For example, one can instantiate a component via:
...
+
 Once the component is thus defined, it can be used in a controller, for example via:
 
     @action('mypage')
     @action.uses(session)
     def mypage():
         c = my_component(id=1)
         return dict(my_component=my_component)
-        
+
```
agavgavi, line 42: Where is c being used here? The line below mentions `my_component(id=1)` which is set to c. Should the return instead be `return dict(my_component=c)` ?

lucadealfaro: Thanks, typo fixed.

lucadealfaro:
I have added sortable fields in tables as well.

mdipierro:
Really really nice. Not sure we should promise backward compatibility on this quite yet as we may want more feedback from users.

lucadealfaro:
Yes, you are right, I think we should add a note saying that this is not fully stable yet. 
But I find these components are shaping up to be nicely useful.

Issue #223, opened 2020-06-25
psycopg2.errors.NotNullViolation: null value in column "id" violates not-null constraint


wasare:
Insert using example form causing error "column "id" violates not-null constraint"
I've coded a new form based on "example form" (create_form action) but on submit cause the error: psycopg2.errors.NotNullViolation: null value in column "id" violates not-null constraint

The form is passing the correct vars data dict with "id: None" and to database as a null value. The PostgreSQL sequence for auto increment the id field was working properly. PostgreSQL waiting a "DEFAULT" value to id generate the new primary key, but receiving null instead. I didn't found a workaround to solve this trouble. Maybe a pydal related issue.

Software info:
py4web 1.20200517.1
pydal 20200531.3
psycopg2-binary 2.8.5
PostgreSQL 11.8 on x86_64-apple-darwin16.7.0, compiled by Apple LLVM version 8.1.0 (clang-802.0.42), 64-bit

wasare:
Tested workaround:  add to models.py for each table => db.table_name.id.writable=False

wasare:
I fixed this issue in utils.form.py i considered the error is not pydal related. If it have another way to fix that please guide me. I will glad to help with it.

PR #225, opened 2020-06-26
fix error: "id" violates not-null constraint in utils.form

wasare:
error found when connected to a postgresql database.
ref: #223 psycopg2.errors.NotNullViolation: null value in column "id" violates not-null constraint

I fixed this issue in utils.form.py i considered the error is not pydal related. If it have another way to fix that please guide me. I will glad to help with it.


wasare:
I fixed this issue in utils.form.py i considered the error is not pydal related. If it have another way to fix that please guide me. I will glad to help with it.

PR #232, opened 2020-07-30
Adding Okta as Oauth providers, adding mandatory state parameter for oauth, extracting user attributes from token

Kkeller83:
Added Okta as an identity provider plugin
Added mandatory state parameter in Oauth flow generated randomly and checked upon IDM callback
Removed fetching of user attributes from idm via userinfo endpoint and changed to extracting user attributes directly from 
oauth access token.

### DIFF
--- a/apps/_scaffold/common.py
+++ b/apps/_scaffold/common.py
@@ -29,7 +29,7 @@ for item in settings.LOGGERS:
 
 # connect to db
 db = DAL(settings.DB_URI,
-         folder=settings.DB_FOLDER, 
+         folder=settings.DB_FOLDER,
          pool_size=settings.DB_POOL_SIZE,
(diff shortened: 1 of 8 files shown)


Kkeller83:
jwt decode without verifying with the key from the identity manager is not best practice, but I have it like this, because it insures that it gets the developer up to speed. I would need to add a documentation or a configuration option next that enables the developer to change the code to enter the JWT secret from the IDM so the token can verified offline,

Kkeller83:
Line 147 in __init.py__ of auth_plugins   : 

res=jwt.decode(token, verify=False)

With the JWT secret which should be used the code should look like this: 

res=jwt.decode(token,'secret', algorithms=['RS256']) (or HS256 depending on the IDMs algorithm used)

mdipierro:
fantastic!

PR #236, opened 2020-08-04
Py4web feature proposal. App Watch.

karbiv:
Py4web provides development server setup with a watchdog lib functionality, but only for the applications python files.
It would be useful to have one decorator that binds a handler for other listed files in any application module:

```py
from py4web.core import app_watch_handler
import sass # https://github.com/sass/libsass-python

@app_watch_handler(
    ["static_dev/sass/all.sass",
     "static_dev/sass/main.sass",
     "static_dev/sass/overrides.sass"])
def sass_compile(changed_files):
    print(changed_files) # for info, files that changed, from a list of watched files above
    ## ...
    compiled_css = sass.compile(filename=filep, include_paths=includes,
                                output_style='compressed')

    dest = os.path.join(app, 'static/css/all.css')
    with open(dest, 'w') as file:
        file.write(compiled)
```
Handlers are executed in Python's import order - `@app_watch_handler`s in application's `__init__.py` executed last.
Paths in decorator are relative to current app's root, without app name(folder name can change).
Those handlers are invoked only when `py4web` dev server is running.
App files with ".py" suffix are ignored by App Watch decorator.

Flask proposes in docs to use WSGI middleware and manual watchdog setup, and there's no notion of applications.
Py4web has watchdog setup ready to use.
For web frameworks it's customary to use `npm` and _gulpfile.js_(or Grunt), _node_modules_ folder with hundreds of dependencies, a separate workflow, Node.js dev server to watch files.

This feature was written as a trial for `py4web` while choosing between Python frameworks for an old site. It looked like Flask required more code just to start.


mdipierro:
This is nice. Give me some time to review it carefully.

karbiv:
Example how to use javascript validation.
`npm`(node.js) in [gulp-jsvalidate](https://www.npmjs.com/package/gulp-jsvalidate) package uses library called `esprima` by Ariya Hidayat.

There's a full python version of `esprima` by Kronuz, [esprima-python](https://github.com/Kronuz/esprima-python).
```py
import esprima

@app_watch_handler(
    ["static/js/index.js",
     "static/js/utils.js",
     "static/js/dbadmin.js"])
def validate_js(changed_files):
    for cf in changed_files:
        print("JS syntax validation: ", cf)
        with open(os.path.abspath(cf)) as code:
            esprima.parseModule(code.read())
```

If some syntax is wrong or with typo - shows an exception in terminal.
Parse time is unnoticeable, for scripts like minified jquery.

mdipierro:
I like this. Can you also include your examples in the documentation?

karbiv:
I can.
It can be added to the bottom of "Creating your first app", if not excessive.
It's a development utility not required for a minimal application, could reside in _project setup recipes_ or something. The book format is not always easily approachable in comparison with URL categories and subsections.

I wrote this feature while checking how `py4web` plays along with [poetry](https://python-poetry.org/) dependency manager(automatic virtual env, if there are any overlapping functionalities). This is a grey area which I thought of addressing in a discussional issue, maybe.

karbiv:
Last paragraph in chapter-02.mm, after a sentence that reads "any JS library that you want because py4web is agnostic to your choice".

PR #238, opened 2020-08-04
Make field 'comments' available to custom forms

jpsteil:
Same as adding the label attribute, this change pulls the comments attribute of a Field into the control and makes it available to custom forms.  I sometimes like to use this as a little helper for the user on what the field is for.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -65,6 +65,7 @@ class FormStyleFactory:
         controls = dict(
             labels=dict(),
             widgets=dict(),
+            comments=dict(),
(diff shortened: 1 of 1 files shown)


mdipierro:
Being the devil's advocate. Do we need this? what is it for? How about placeholders instead?

qualityliquidfeeds:
The comment attribute is available SQLFORM.factory in web2py.  My intention was to use it the same way in py4web.  I agree placeholders may be a better implementation on the HTML site, but I'd still want to pass the string value from the Form().  I'm open to alternatives...

PR #242, opened 2020-08-10
Bugfix Oauth statechecker

Kkeller83:
Fix bug always returning "401 not authorized" when using Oauth, because passedstate variable was never defined. Sorry, I did not merge my code corretly before. Just went through a whole other test cycle and now it all works as expected and as it was in my orginial development code. Next time I will directly merge the code from my development environment in order to avoid any "code replication" errors, by missing to add some of the changes into the git commit that I did in  my development environment.

### DIFF
--- a/py4web/utils/auth_plugins/__init__.py
+++ b/py4web/utils/auth_plugins/__init__.py
@@ -122,6 +122,8 @@ class OAuth2(SSO):
         )
         scope = self.parameters.get("scope")
         state = self.state_generator()
+        global passedstate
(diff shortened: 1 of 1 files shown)


mdipierro:
put passed_state in the plugin.parameters. Let me know if works for you

Kkeller83:
Had a look, yeah this is more elegant.

Should work fine.

Thanks!

Am So., 16. Aug. 2020 um 19:03 Uhr schrieb mdipierro <
notifications@github.com>:

PR #244, opened 2020-08-11
Adjusted Fontawesome 5 to CSS over JS due to SVG issues with icons.

agavgavi:
With Fontawesome 5, icons have translated over to SVG and are unwrapped in the DOM when using the `.js` version currently in `layout.html`. This leads to some issues updating classes with Vue, you can no longer follow the same methodology as before. This led to Luca's Starrater component to break in multiple ways.

Currently the best was I've found is to use the WebCSS version of fontawesome instead of the SVG/JS version which works, but if any example apps require the version Massimo already has in `layout.html` when he updated to Font Awesome 5, this would not work.

Another solution would be to instead of automatically including font-awesome in `layout.html`, that would be delegated to a developer inserting it into the `block page_head`. That way developers could select between if they want CSS or JS.

Since @lucadealfaro is the creator of these components, he may have a better solution in mind, but this is just an attempt to fix it and get it to work in the current version of py4web.

### DIFF
--- a/apps/examples/static/components/starrater/starrater.html
+++ b/apps/examples/static/components/starrater/starrater.html
@@ -2,7 +2,7 @@
     <span v-on:mouseover="stars_over(idx)"
           v-on:click="set_stars(idx)"
           v-for="idx in star_indices">
-        <i v-if="idx <= num_stars_display" class="fa fa-star"></i>
-        <i v-if="idx > num_stars_display" class="fa fa-star-o"></i>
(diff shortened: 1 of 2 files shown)


lucadealfaro:
@mdipierro , any reason we are not using font awesome 4.7, which does not require (and thus break) js?

mdipierro:
Ignorance. I will move to 4.7.

agavgavi:
Perfect, I'll close this PR then.

PR #251, opened 2020-08-14
ThreadSafe pydal patch

valq7711:
There is a bug in case of usage [Table-inheritance](http://www.web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#Table-inheritance)
The reason:
"copy.copy" makes a shallow copy,
but Field ThreadSafe-attributes ("readable", "writable", "default", "update", "requires") are class-level as they are [patched](https://github.com/web2py/py4web/blob/2865f3929ab1dd837d3f8566d6a9a3d6621c9c0e/py4web/core.py#L289-L292) by py4web , so:
no copy -> no attr in ICECUBE for the fresh one -> gevent-error on try to access to any of ThreadSafe-attributes

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -6,6 +6,7 @@
 import argparse
 import cgitb
 import code
+import copy
 import datetime
 import functools
 import importlib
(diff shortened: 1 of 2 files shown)


mdipierro:
I think this logic should eventually be moved to pydal. But thanks this is excellent.

valq7711:
Yes, I also think that pybal is better place, but perfection requires all the time in the world :disappointed: 
Thank you for another great framework! Glad to help

PR #253, opened 2020-08-17
Fix set-password in README

wymored:
README.md file has a small typo error :  

     py4web set_password 

instead of 

     py4web set-password 

this pull request fix both

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ git clone https://github.com/web2py/py4web.git
 cd py4web
 python3 -m pip install -r requirements.txt
 ./py4web.py setup apps
-./py4web.py set_password
+./py4web.py set-password
 ./py4web.py run apps
 open http://localhost:8000/todo/index
 ```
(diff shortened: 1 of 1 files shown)


nicozanf:
Hi, there are different point of view - I think that the current standard command is now `set_password` ....

wymored:
> 
> 
> Hi, there are different point of view - I think that the current standard command is now `set_password` ....
I used the github source code, is result is below:

image

nicozanf:
We are both right ;-)


On the code, there is indeed an undescore (see [here](https://github.com/web2py/py4web/blob/18a09f9b83f20b5538847e9079ecbd1aa21a84af/py4web/core.py#L1422) ).
But we've faced a _new feature_ of Click version 7.0 that silently changes it to a dash:
"The name of the command defaults to the name of the function with underscores replaced by dashes. If you want to change that, you can pass the intended name as the first argument." (on [their docs](https://click.palletsprojects.com/en/7.x/api/)).

So, let's ask @mdipierro : do you want to have **set-password** or **set_password** ?

Then we can change the code and docs accordingly 

Nico

PR #255, opened 2020-08-17
fix auth/api

valq7711:
I met a bug  - `GET auth/api/logout` - returns success, but nothing is performed (user is still in session). I suppose this is not the only bug like this,  because there is not path-method validation. Therefor I rewrote auth/api in more reliable/readable style. Now, all api-handlers in the `auth.api` that is dict like `{'profile': {'GET': get_profile_handler, 'POST':  post_profile_handler,}, ...}` which is generated by `auth.make_api()` at `auth.__init__()`.
Also, I have some questions about login/logout logic  - see comments in code.
And one more note about logic:
As I see, in case of any errors api returns code 200 with actual code and error-message in the response. As for me (given that this is **api**) it is overhead and not convenient. I think api should response with actual code and processed in this common way:
```javascript
        axios.post('../auth/api/register', form)
            .then(function(res){self.go('registered')})
            .catch(function(err){
                var res = err.response;
                if(res.data.errors) self.errors = res.data.errors;
                else alert(res.data.message);
            });
```
instead of:
https://github.com/web2py/py4web/blob/75cbe6229e8530d578c41da45c009580dc6da86f/apps/_scaffold/static/components/auth.js#L69-L73

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -406,10 +406,13 @@ class Auth(Fixture):
 
             @mount('POST')
             def login(vars, *a, **v):
+                username, password = vars.get("email"), vars.get("password")
+                if not all(isinstance(_, str) for _ in [username, password]):
+                    return self._error("Invalid Credentials")
+
                 # Prioritize PAM or LDAP logins if enabled
                 if "pam" in self.plugins or "ldap" in self.plugins:
                     plugin_name = "pam" if "pam" in self.plugins else "ldap"
-                    username, password = vars.get("email"), vars.get("password")
                     check = self.plugins[plugin_name].check_credentials(
                         username, password
                     )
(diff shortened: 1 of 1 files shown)


mdipierro:
I love what you have done but I think on logout we will have to clear the session. It is not sufficient to remove the user. The reason is that the session may leak info. Can you make that change?

valq7711:
@mdipierro So can I change `auth/api` response to:
```python
            data["status"] = data.get("status", "success")
            response.status = data["code"] = data.get("code", 200)
            return data
``` 
and fix auth.js to:
```javascript
axios.post('../auth/api/register', form)
            .then(function(res) {self.go('registered')})
            .catch(function(err){
                var res = err.response;
                if(res.data.errors) self.errors = res.data.errors;
                else alert(res.data.message);
            });
```
???

PR #256, opened 2020-08-17
replace everywhere 'set-password' with 'set_password'

nicozanf:


nicozanf:
As noted on PR #253 , the module Click (from version 7) silently changes the command set_password to set-password. This patch forces the use of **set_password** and changes the docs / scripts accordingly.

Approve it only if **set_password** is the right command's name.

PR #259, opened 2020-08-19
fix flash

valq7711:
`utils.flash([[=XML("hello world")]]) -> utils.flash(hello world)`

### DIFF
--- a/apps/_scaffold/templates/layout.html
+++ b/apps/_scaffold/templates/layout.html
@@ -58,6 +58,6 @@
   <script src="js/axios.min.js"></script>
   <script src="js/vue.min.js"></script>
   <script src="js/utils.js"></script>
-  [[if globals().get('flash'):]]<script>utils.flash([[=XML(flash)]]);</script>[[pass]]
+  [[if globals().get('flash'):]]<script>utils.flash([[=XML("\"%s\"" % flash) if isinstance(flash, str) else XML(flash)]]);</script>[[pass]]
   [[block page_scripts]]<!-- individual pages can add scripts here -->[[end]]
 </html>


apps/_scaffold/static/js/utils.js
```diff
@@ -245,10 +245,13 @@ utils.handle_flash = function() {
         return function (event) { 
             var id = 'notification-{0}'.format([element.dataset.counter]);
             element.dataset.counter = parseInt(element.dataset.counter) + 1;
+            // event.detail can be a string, but we need hash like {message: ..., class: ...}
+            var e_detail = event.detail; // event.detail is readonly!
+            if (typeof e_detail == "string"){e_detail = {message: e_detail}}
```
mdipierro, line 250: I may be ok with this

apps/_scaffold/templates/layout.html
```diff
@@ -58,6 +58,6 @@
   <script src="js/axios.min.js"></script>
   <script src="js/vue.min.js"></script>
   <script src="js/utils.js"></script>
-  [[if globals().get('flash'):]]<script>utils.flash([[=XML(flash)]]);</script>[[pass]]
+  [[if globals().get('flash'):]]<script>utils.flash([[=XML("\"%s\"" % flash) if isinstance(flash, str) else XML(flash)]]);</script>[[pass]]
```
mdipierro, line 61: Not fine with this. It was cryptic enough as it was :-(

apps/_scaffold/static/js/utils.js
valq7711: Yes, it looks not perfect, but `utils.flash([[=XML("hello world")]])` turns into `utils.flash(hello world)`, that causes an js-error

apps/_scaffold/templates/layout.html
valq7711: Yes, it looks not perfect, but `utils.flash([[=XML("hello world")]])` turns into `utils.flash(hello world)`, that causes an js-error

PR #267, opened 2020-08-24
Form adjustments

agavgavi:
When working with custom forms, there was an issue with upload fields not appearing in the self.vars of a custom form. So if a developer makes a custom form not  attached to a database but a list with a field of type `upload`, when submitting the form will not recognize the field. This was fixed with the assistance of people in the py4web google group [here ](https://groups.google.com/g/py4web/c/APh1hPl4vPE) by adding `self.vars.update(validated_vars)`, inserting all validated variables to the vars dict.

In addition, added a formstyle function like `FormStyleBulma` for Bootstrap known as `FormStyleBootStrap` and readjusted the file uploader field of the form to be laid out on 2 lines instead of 1 long one.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -2,6 +2,7 @@ import json
 import jwt
 import time
 import uuid
+import copy
 from py4web import request, Session
 from pydal.validators import Validator
 from yatl.helpers import (
@@ -405,7 +406,8 @@ class Form(object):
             if self.record:
                 self.vars = self._read_vars_from_record(table)
         else:
-            post_vars = self.vars = request.forms
+            post_vars = request.forms
+            self.vars = copy.deepcopy(request.forms)
             self.submitted = True
             process = False
 
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -295,7 +295,7 @@ def FormStyleBulma(table, vars, errors, readonly, deletable):
     return FormStyleDefault(table, vars, errors, readonly, deletable, classes)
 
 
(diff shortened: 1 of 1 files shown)


valq7711:
some thoughts:

I think extra validation should go after base validation and after form.vars update https://github.com/web2py/py4web/blob/4f6e297d5ceb150d786c6614550a8ba414022d08/py4web/utils/form.py#L448-L449

Since we modify self.vars, it might be better to make a copy from `request.forms`
https://github.com/web2py/py4web/blob/4f6e297d5ceb150d786c6614550a8ba414022d08/py4web/utils/form.py#L408

agavgavi:
I agree with you with moving the extra validation, but I think I'm not understanding your second point. Are you saying to make a deep copy of `self.vars` instead of just referencing `request.forms`?

agavgavi:
Nevermind I see and agree, deepcopy on self.vars make sense. We don't need to worry about that with post_vars though it seems since all that is being used for is retrieval instead of modification.

py4web/utils/form.py
```diff
@@ -280,6 +295,29 @@ def FormStyleBulma(table, vars, errors, readonly, deletable):
     return FormStyleDefault(table, vars, errors, readonly, deletable, classes)
 
 
+def FormStyleBootStrap(table, vars, errors, readonly, deletable):
```
mdipierro, line 298: Maybe FormStyleBootStrap4? We run into this issue for web2py

agavgavi: makes sense, has been changed.

PR #269, opened 2020-08-26
handle missing _class in control

jpsteil:
_class may sometimes not be in control


mdipierro:
thanks but this was fixed. I did not see your patch in time.

PR #283, opened 2020-09-11
User can set formstyle on auth forms

jpsteil:
Allow the user to set the formstyle for all auth forms globally.  Default to FormStyleDefault 

In common.py you can set:
auth = Auth(session, db, define_tables=False, formstyle=FormStyleBulma)

to default all auth forms to Bulma, you just need to import FormStyleBulma in common.py.

-Set login form to use proper fields/labels based on auth.use_username value
-Show/hide Sign Up / Lost Password only if allowed in allowed_actions

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -10,7 +10,7 @@ import uuid
 
 from py4web import redirect, request, response, abort, URL, action, Field, HTTP
 from py4web.core import Fixture, Template, Flash, REGEX_APPJSON
-from py4web.utils.form import Form
+from py4web.utils.form import Form, FormStyleDefault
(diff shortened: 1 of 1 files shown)


mdipierro:
Let's discuss this tomorrow morning. I would prefer a slightly different solution to avoid the mess we got into with web2py. I would prefer that Auth() does not know about type but only AuthForms() do. I would like the latter to be passed to the former and the style be set by overriding AuthForms. Or something alike. Thougths?

jpsteil:
I will be on the chat tomorrow so we can discuss.  I am in favor of
implementing however you think it makes sense.
-Jim

mdipierro:
closing because already addressed

PR #287, opened 2020-09-14
New Grid

jpsteil:
Replaced grid.py in utils with new grid
Updated html grid in examples app
Added new font-awesome cdn link in examples and _scaffold.


jpsteil:
Massimo - I think we need to chat a bit about the usage of params.  I think I used it incorrectly in my implementation.  I'm using it in grid.py to send common parameters for the app.  I may need some coaching on how to set it up like auth so we can show consistent usage and examples of params.

mdipierro:
The way I have been using it is this

from py4web.utils.param import Param

class Something:
    def __init__(self):
        self.param = Param(a=1, b=2, c=3)

the use:
self.param.a
self.param.b
self.param.c

we want to put into param values into the constructor that can be changed by the users later.
param values should only be constant input values that affect the behavior of the object.

In your case:

grid =  Grid()
grid.param.XYZ = …

Not sure what those parameters should be.
Can chat tonight or tomorrow night if works for you. The later the better.

qualityliquidfeeds:
I need to look at your auth a bit more first.  

At this time, Grid works like SQLFORM.grid.  You call it with the parms needed and it does something.  I need to look at refactoring a bit so you can call it in common.py like you do with auth, set the default parms and then call grid.Grid() or something each time you want to use it in your app.  I'll give it a go today and report back.

mdipierro:
No. I think it is fine as it is. Defining one grid in common and then
changing attributes risks not being thread safe. Auth is a fixture. Form
and grid are not and I think better this way.

jpsteil:
I have updated to using Param more effectively, will be sending new PR this afternoon.  I'd appreciate your critique as you have time.

jpsteil:
This PR now has my latest changes.  I thought I had to create a new PR, but my latest changes are now included in this one.

mdipierro:
Will try look at this tonight. May slip to tomorrow

jpsteil:
No rush. I'll keep making changes and add to the pull request.

Maybe get some time tomorrow to look at email auth again.

-Jim

mdipierro:
I am merging this as is. Will make some minor changes after merge. Good work!

PR #297, opened 2020-09-28
Refactored grid signature and implement Param class

jpsteil:


jpsteil:
I'm hoping this will solidify the grid signature for now.  Added parameter for search_queries, but it doesn't do anything yet.

PR #298, opened 2020-10-01
fixed FB login for some cases. Handle corner case in session/user 

tolgakonik:
see comments on code. The first one on session is a minor defense. It fixes a problem that I rarely experience.
The second on FB login is fallback to previous approach when a recently implemented approach is not working. (happened in my case).


Kkeller83:
Looks good! 
It will however not handle all cases. 
For social IDPs you only get an access token.
For "enterprise" IDPs such as Okta, Akamai, Auth0 etc. you will usually get a set of tokens (id_token,access_token and refresh token). Sometimes in opaque format for access and refresh tokens at least and sometimes in JWT format. 
So even with your added code, there will most likely often cases where you ll have to modify this to suit your IDP needs. 
But I really like your pull request, as it is a good step forward to cover more cases out of the box.

PR #305, opened 2020-10-04
Use Colorama module for ANSI printout

nicozanf:
The colorama module is automatically used by Click if present. This is highly compatible (especially for Windows that otherwise has many problems with ANSI) and now we can simply and effectively use colors! Tested with Win , Mac, Linux + python 3.7 and 3.8


mdipierro:
Love this!

PR #315, opened 2020-10-15
Formatting changes - fix initial sorter icon missing

jpsteil:
Grid Cell Formatting
when building table cell for the grid include the type-specific formats as well as the default grid-td

Formatters
change the lambda function for booleans

Remove Unused __init__ parameters

Sort Order
Sorter icons didn't display if no sort order was included on the url query string.  Added code to look at the grid instance param.orderby to determine sort order if the orderby url query was empty

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -62,6 +62,9 @@ class GridClassStyle:
         "grid-sorter-icon-down": "grid-sort-icon-down fas fa-sort-down",
         "grid-th-action-button": "grid-col-action-button",
         "grid-td-action-button": "grid-col-action-button",
+        "grid-tr": "",
+        "grid-th": "",
+        "grid-td": "",
         "grid-details-button": "grid-details-button info",
         "grid-edit-button": "grid-edit-button info",
         "grid-delete-button": "grid-delete-button info",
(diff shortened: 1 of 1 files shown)


py4web/utils/grid.py
```diff
@@ -668,6 +672,17 @@ def render_table_header(self):
...
+        if sort_order:
+            sort_order = [sort_order]
+        else:
+            sort_order = []
+            for x in self.param.orderby:
+                op = ''
+                if hasattr(x, 'op') and x.op and x.op.__name__ == 'invert':
+                    x = x.first
+                    op = '~'
+                sort_order.append("%s%s.%s" % (op, x.tablename, x.name))
```
mdipierro, line 684: Why append and not inert(0,...)?

jpsteil: I don't have a good reason.  This section of code something (I think) Val had suggested when I had a problem earlier with recognizing DESC.  I have no problem if you'd like to change it.

mdipierro: i will merge everything but this because unfortunately does not work as it unless we restrict orderby to expressions that are representable as strings. Need to think more about it. Notice the user can always do: if not request.query.get('orderby'): sequet.query['orderby'] = "~tablename.fieldname".

jpsteil: Sounds good.  I will try to find a better solution.

PR #339, opened 2020-11-13
improve URL

valq7711:
allow to map domains to apps
example of usage:
```nginx
 map $http_host $appname {
       hostnames;
       foo.*   /some;  # foo.some.org  mapped to '/some' app
       .bar.com   /other;
 }
 server {
        listen       80;
        listen       [::]:80;
        server_name  foo.baz.org  www.bar.com  .example.*;

    location / {
        proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header   Host $http_host;
        proxy_set_header   X-Real-IP $remote_addr;
        proxy_set_header   X-Forwarded-Host $server_name;
        proxy_redirect off;
        # for app that is not mapped we should just unset X-Py4web-Appname
        location /not_mapped_app/ {
                proxy_set_header X-Py4web-Appname "";
                proxy_pass http://unix:/run/gunicorn.sock;
        }
       # if this header is set and not empty then app is mapped => URL() should work with use_appname= False
        proxy_set_header X-Py4web-Appname $appname;

        # magic here  - we just add $appname to uri 
        # note that .example.* is not mapped and in this case $appname will be ""
        # so for  example.com  all apps  will be served in regular mode! i.e. as http://example.com/some 
        rewrite ^ $appname$request_uri break;
        proxy_pass http://unix:/run/gunicorn.sock;
    }
}


```

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -580,7 +580,7 @@ def URL(
     hash=None,
     scheme=False,
     signer=None,
-    use_appname=True,
+    use_appname=None,
     static_version=None,
 ):
     """
@@ -591,6 +591,9 @@ def URL(
     URL('a','b',vars=dict(x=1),scheme='https') -> https://{domain}/{script_name?}/{app_name}/a/b?x=1
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
+    if use_appname is None:
+        use_appname = request.headers.get("x-py4web-appname")
+        use_appname = True if use_appname is None else not use_appname
     script_name = (
         request.environ.get("HTTP_X_SCRIPT_NAME", "")
         or request.environ.get("SCRIPT_NAME", "")
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -1115,8 +1115,18 @@ def error_page(code, button_text=None, href="#", color=None, message=None):
 
 @bottle.error(404)
 def error404(error):
-    guess_app_name = request.path.split("/")[1]
-    return error_page(404, button_text=guess_app_name, href="/" + guess_app_name)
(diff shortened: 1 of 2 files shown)


py4web/core.py
```diff
@@ -591,6 +591,9 @@ def URL(
     URL('a','b',vars=dict(x=1),scheme='https') -> https://{domain}/{script_name?}/{app_name}/a/b?x=1
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
+    if use_appname is None:
+        use_appname = request.headers.get("x-py4web-appname")
+        use_appname = True if use_appname is None else not use_appname
```
mdipierro, line 596: Do not understand this. "x-py4web-appname": "false" would evaluate to true?

valq7711: Yes, it looks horrible and would be better to have `proxy_set_header x-py4web-use-appname !$appname`, but nginx doesnt support these things and it requires if-block + yet another var

valq7711: Just in case:
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

misl6:
For being sure that this one works out-of-the-box also for apps that uses the auth module, how it sounds to change also this param to be `None` by default?

https://github.com/web2py/py4web/blob/c1dcf19a837a3835684d8f1e86c051a500499605/py4web/utils/auth.py#L145

valq7711:
@mdipierro    the logic is absolutely the same as for x-script-name: the proxy server adds a prefix to the uri and passes information about this to wsgi

valq7711:
@mdipierro  Why is `use_appname` needed at all?

PR #346, opened 2020-11-16
handle translations in grid titles

jpsteil:
When converting text to a title, consider it may be a translation.


jpsteil:
Added another commit to change new-app cli command to new_app

PR #347, opened 2020-11-16
Add missing DOCS files under _documentation app

nicozanf:
There is a .gitignore setting that is normally preventing to upload the generated documentation, if not forced.

I've then changed docs/readme.MD adding:

NOTE: if you use a local git repository and you wish to update the documentation to the remote repository, too, you have to force the changes because they are not normally allowed by .gitignore. In order to do it, you have to issue a 'git add --force apps/_documentation/*' from the py4web folder

And I've modified updateDocs.sh in order to behave correctly if PDF or EPUB files are not present.


mdipierro:
I am confused. I thought the sources lived under py4web/docs. Shouldn't the makefile just put html files under apps/_documentations/static/ without committing them?

agavgavi:
The reason for including the docs in this folder is that by default a user wouldn't get a documentation app if they clone from source. The app would 404 and unless they are apart of the py4web group, they would have no idea how to install the apps. The solution is to either give detailed instructions on the installation README, or every time there is a version release (like when its updated on pip) make sure to run the Makefile so that the doc is somewhat updated?

The discussion of this can be seen here: https://groups.google.com/g/py4web/c/yBnX7N2LDfc/m/-0gduf0OAwAJ

And In addition the UpdateDoc.sh script had an issue that needs to be fixed in that it won't update if there is no pdf or epub in the folder.

nicozanf:
Exactly.
The sources (and settings + programs) are under /docs.
At every significant change someone should manually run the program inside
in order to get the updated HTML/PDF/EPUB files under
/apps/_documentation/static/, so the final user will get an updated
documentation locally installed.
The same files should finally be copied to
https://py4web.com/_documentation/static

This is what I've setup. Is it right?

Nico

Il giorno gio 19 nov 2020 alle ore 07:35 Andrew Gavgavian <
notifications@github.com> ha scritto:

mdipierro:
Sounds right. I will review tomorrow. Thanks.

PR #354, opened 2020-11-23
Add new_app docstrings

nicozanf:
And name it right (not new-app as converted automatically but new_app, as it was originally intended)


nicozanf:
I'm redoing it with just the docstring ...

PR #366, opened 2020-11-29
Update rendered docs

nicozanf:
Reflects PR #365, to be manually pushed to https://py4web.com/_documentation/static , too


mdipierro:
I am still not convinced we need to commit rendered docs. :-)

PR #368, opened 2020-12-01
Fixes next in redirects

misl6:
This PR fixes `next` in redirects if `use_appname` is None/False or `X-Py4web-Appname` is set

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -65,15 +65,18 @@ class AuthEnforcer(Fixture):
         else redirects to page"""
         if re.search(REGEX_APPJSON, request.headers.get("accept", "")):
             abort(403)
-        redirect_next = request.fullpath
-        if request.query_string:
-            redirect_next = redirect_next + "?{}".format(request.query_string)
+        noappname_req_path = request.fullpath.split(f"/{request.app_name}")[-1]
+        next_url = URL(
+            noappname_req_path,
+            vars=request.query,
+            use_appname=self.auth.param.use_appname_in_redirects,
+        )
         self.auth.flash.set(message)
         redirect(
             URL(
                 self.auth.route,
                 page,
-                vars=dict(next=redirect_next),
+                vars=dict(next=next_url),
                 use_appname=self.auth.param.use_appname_in_redirects,
             )
         )


qualityliquidfeeds:
FWIW - Val has been helping me come up with a monkey patch to fix request.fullpath in the event that X-Py4web-Appname is set.  Will be submitting a PR later today.  I'm not saying this PR isn't valid.  Just that there is another one coming soon to handle this same situation, but fixes request.fullpath anywhere it is called.

jpsteil:
FYI - that was me in the previous response, logged in to the wrong github account to reply at first.

misl6:
Great!
I needed a quick fix for some apps that are behind a nginx reverse proxy where this specific case was creating some issues to the final user, but I'm happy to hear that there's something evolving that fixes all the cases.

valq7711:
@misl6 you can try:

```python
# core.py

... # imports
import threadsafevariable

########################################################################################
# fix request.fullpath for the case of domain mapping to app (request.url will be autofixed, since it is based on request.fullpath
#########################################################################################
def monkey_patch_bottle():
    urljoin = urllib.parse.urljoin
    @property
    def fullpath(self):
        appname = self.get_header('x-py4web-appname', '/')
        return urljoin(self.script_name, self.path[len(appname):])
    setattr(bottle.BaseRequest, 'fullpath', fullpath)
monkey_patch_bottle()
############################################################################

bottle.BaseRequest.MEMFILE_MAX = 16 * 1024 * 1024
```

valq7711:
just in case I made some fixes here https://github.com/web2py/py4web/pull/339
it is
```
     ...
     proxy_set_header   X-Forwarded-Host $http_host;   # not $server_name,  since we want mapping
     ...
     rewrite ^ $appname$uri break;   # not $request_uri
    
```

misl6:
> @misl6 you can try:
> 
> ```python
> # core.py
> 
> ... # imports
> import threadsafevariable
> 
> ########################################################################################
> # fix request.fullpath for the case of domain mapping to app (request.url will be autofixed, since it is based on request.fullpath
> #########################################################################################
> def monkey_patch_bottle():
>     urljoin = urllib.parse.urljoin
>     @property
>     def fullpath(self):
>         appname = self.get_header('x-py4web-appname', '/')
>         return urljoin(self.script_name, self.path[len(appname):])
>     setattr(bottle.BaseRequest, 'fullpath', fullpath)
> monkey_patch_bottle()
> ############################################################################
> 
> bottle.BaseRequest.MEMFILE_MAX = 16 * 1024 * 1024
> ```

@valq7711 I tested it in a staging environment and it worked flawlessly.

I have a similar nginx configuration to the one you mentioned.

valq7711:
Hi guys! I have an idea!
Serving app under subfolder is fully opposite the case of mapping domain to app, hence these cases never  mix!
But if it is so,  we can utilize `script_name` header instead of custom header (x-py4web-appname) but in special way:  
To indicate that we under mapping (not under subfolder) we can add some prefix to `$appname` that is never used as part of uri, i.e. we can do the trick like:
`proxy_set_header Script-Name !:$appname; `
and adjust monkey patch:
```python
def monkey_patch_bottle():
    urljoin = urllib.parse.urljoin
    @property
    def fullpath(self):
        script_name = self.script_name
        path = self.path
        if script_name.startswith('!:'):
            path = path[len(script_name)-2:] # -2 due to '!:'
            script_name = '/'
        return urljoin(script_name, path)
    setattr(bottle.BaseRequest, 'fullpath', fullpath)
monkey_patch_bottle()
```
also it  requires to fix py4web URL() 

What do you think?

py4web/utils/auth.py
```diff
@@ -65,15 +65,18 @@ def abort_or_redirect(self, page, message=""):
         else redirects to page"""
         if re.search(REGEX_APPJSON, request.headers.get("accept", "")):
             abort(403)
-        redirect_next = request.fullpath
-        if request.query_string:
-            redirect_next = redirect_next + "?{}".format(request.query_string)
+        noappname_req_path = request.fullpath.split(f"/{request.app_name}")[-1]
```
mdipierro, line 68: This will break tings if there is an action with the same name as the app_name

valq7711: Yes, there is the same issue now with _default app, no?

jpsteil: My recollection is that we didn't need the 'next' fix if we have Val's bottle monkey patch fix.  Then next works fine.

valq7711: Yes, it is so.

valq7711: this is the opposite case where the application is served under subfolder (script_name header). In that case fullpath is script_name + path

valq7711: but ambiguous routes can still occur and as far as I understand this problem cannot be avoided

valq7711: the only way is to have strange names for apps like `_whatever` or `whatever_app` that are never used as action name

jpsteil: For what its worth, the monkey patch solution is working really well for me in both dev and production environments.

jpsteil:
Can you check to see whether or not this has been addressed with the recently merged PRs.  I don't believe this PR is necessary any longer.  If not necessary we can close.

PR #379, opened 2020-12-14
Fix #360 : form update sets fields not present in form to null

### LINKED ISSUE #360: form.custom resets fields not presents in form
Having a model with field a, b and using form.custom to edit only field a, field b content will be cancelled.

I know that I should use .writable = False but this behaviour could be harmful.

in models.py

```
db.define_table('test',Field('a'), Field('b'))
 
```

in controller.py
```
@action("edit/<id>", method=["POST", "GET"])
@action.uses(session, db, auth, "edit.html")
def edit(id=None):
    form = Form(db.test, id, deletable=False)
    return dict(form=form)       
```

in edit.html
```
    [[ =form.custom['begin'] ]]
    [[ =form.custom['labels']['a'] ]]
    [[ =form.custom['widgets']['a'] ]]
    [[ =form.custom['submit'] ]]
    [[ =form.custom['end'] ]]

```

mbelletti:
Iterate post_vars instead table.fields

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -455,8 +455,9 @@ class Form(object):
                 record_id = self.record and self.record.get("id")
                 if not post_vars.get("_delete"):
                     validated_vars = {}
-                    for fn in [k for k in post_vars.keys() if not k in ['_formname', '_formkey']]:
-                        field =  self.table[fn]                
+                    for field in self.table:
+                        if not field.name in post_vars: 
+                            continue
                         if field.writable and field.readable and field.type != "id":
                             original_value = post_vars.getall(field.name)
                             if (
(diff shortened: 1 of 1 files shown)


py4web/utils/form.py
```diff
@@ -455,7 +455,8 @@ def __init__(
                 record_id = self.record and self.record.get("id")
                 if not post_vars.get("_delete"):
                     validated_vars = {}
-                    for field in self.table:
+                    for fn in [k for k in post_vars.keys() if not k in ['_formname', '_formkey']]:
+                        field =  self.table[fn]
```
mdipierro, line 459: we should do the opposite.
```
for field in self.table:
    if not field.name in post_vars: continue
```
else field =  self.table[fn]    may result in an unhandled error or a vulnerability.

mbelletti: Done. Thank. you

mbelletti: I don't' know the workflow to fix a PR I hope that is the right way.

PR #381, opened 2020-12-15
Honor 'represent' on columns when displaying in grid

jpsteil:
honor the pydal represent keyword when displaying the table in grid

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -738,9 +738,15 @@ class Grid:
         :return:
         """
         if self.use_tablename:
-            field_value = row[field.tablename][field.name]
+            if 'represent' in field and field.represent:
+                field_value = self.db[field.tablename][field.name].represent(row[field.tablename][field.name])
+            else:
+                field_value = row[field.tablename][field.name]
         else:
-            field_value = row[field.name]
+            if 'represent' in field and field.represent:
+                field_value = self.db[field.tablename][field.name].represent(row[field.name])
+            else:
+                field_value = row[field.name]
         key = "%s.%s" % (field.tablename, field.name)
         formatter = (
             self.formatters.get(key)


py4web/utils/grid.py
```diff
@@ -738,9 +738,15 @@ def render_field(self, row, field, field_index):
...
+            if 'represent' in field and field.represent:
+                field_value = self.db[field.tablename][field.name].represent(row[field.tablename][field.name])
+            else:
+                field_value = row[field.tablename][field.name]
         else:
-            field_value = row[field.name]
+            if 'represent' in field and field.represent:
+                field_value = self.db[field.tablename][field.name].represent(row[field.name])
+            else:
+                field_value = row[field.name]
```
mdipierro, line 749: Instead of all of this I would do
```
        if self.use_tablename:
            field_value = row[field.tablename][field.name]
       else:
            field_value = row[field.name]
       if hasattr(field, 'represent'):
            field_value = field.represent(field_value)
```

mdipierro: Anyway, I am unsure about using represent simply because we have the formatter which is based on the type of field. I am afraid we may end up having conflits beteween represent and formatter. Also represent in py4web is not thread safe. I think we should deprecate it.

jpsteil: Would you be open to supporting represent if we had clear documentation that let people know that it is not thread-safe?

I know there are a few attributes of a field that are thread-safe.  I would like for us to have it clearly stated in the doc which are and which aren't.  

Yes, we have the formatters and they are great, but they have to be specified for each grid invocation.  I believe that supporting the represent attribute is a nice way to set it site-wide.  Formatters can then be used as needed to fine-tune or override.

jpsteil:
Discussed with Massimo today and decided we are going to NOT support represent in grid.  If you want represent-like behaviour you should extend grid and put in the formatters that you want.  I will update grid documentation to reflect this.

PR #385, opened 2020-12-19
Handle left join with no fields or search query/form - fix typo

jpsteil:
Handle something like this.  Without this update the grid was thinking it had to provide the tablename in queries, but without specifying fields from both tables, it didn't accept a tablename.  (Lines 777 and 781)

```
grid = Grid(path,
            query=reduce(lambda a, b: (a & b), [db.links_integr.id > 0]),
            left=[db.nodes_integr.on(db.links_integr.source == db.nodes_integr.id)])
```

The rest of the changes are all typo fixes where 'seach' should be 'search'


mdipierro:
ahhhhhhhhhh!

klausz:
Tested, works.

Issue #390, opened 2020-12-22
error in utils/form.py


ndegroot:
On line 536 I get an error:

'str' object has no attribute 'decode'. 

as jwt.encode() returns a string (maybe it used to be bytes in earlier version?) as stated on:
https://pyjwt.readthedocs.io/en/stable/api.html 

```
536:        self.formkey = jwt.encode(payload, self._get_key(), algorithm="HS256").decode(
            "utf-8"

```
Easy fix: just remove .decode("utf8"). I can make a pull request, maybe first add some tests for form.py?

PR #391, opened 2020-12-23
test added for form issue #390 fixed

ndegroot:
This should fix issue #390 test added (test_form.py) Could be expanded.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -535,9 +535,7 @@ class Form(object):
         payload = {"ts": str(time.time())}
         if self.lifespan is not None:
             payload["exp"] = time.time() + self.lifespan
-        self.formkey = jwt.encode(payload, self._get_key(), algorithm="HS256").decode(
-            "utf-8"
-        )
+        self.formkey = jwt.encode(payload, self._get_key(), algorithm="HS256")
 
     def _verify_form(self, post_vars):
         """Verifies the csrf signature and form name."""
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -1,0 +1,14 @@
+import os
+import unittest
+from py4web import response, Field
+from py4web.utils.form import Form
+
+
+class FormTest(unittest.TestCase):
+    def test_form(self):
+        table = [Field("name")]
+        form_name = 'testing_form'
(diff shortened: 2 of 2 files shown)


py4web/utils/form.py
```diff
@@ -535,9 +535,7 @@ def _sign_form(self):
         payload = {"ts": str(time.time())}
         if self.lifespan is not None:
             payload["exp"] = time.time() + self.lifespan
-        self.formkey = jwt.encode(payload, self._get_key(), algorithm="HS256").decode(
-            "utf-8"
-        )
+        self.formkey = jwt.encode(payload, self._get_key(), algorithm="HS256")
```
mdipierro, line 538: Unfortunately this works in the test below but results in the wrong string being sent in the actual forms. For example, if you try login and explore the multipart form sent from browser to server you will get:

```
-----------------------------178483193414293891411069400419
Content-Disposition: form-data; name="_formkey"

b'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ0cyI6IjE2MDg3Nzc1OTkuOTc4MjAyNiJ9.PhS9K39Rk-cBEqaAmMDhUztTePd0gpGK4-RdKAfYaPk'
-----------------------------178483193414293891411069400419--
```

I recommend leaving this line as is.

tests/test_form.py
```diff
@@ -0,0 +1,14 @@
...
+import unittest
+from py4web import response, Field
+from py4web.utils.form import Form
+
+
+class FormTest(unittest.TestCase):
+    def test_form(self):
+        table = [Field("name")]
+        form_name = 'testing_form'
+        f = Form(table,form_name=form_name)
```
mdipierro, line 11: This test fails for me because Form still tries parse a request object that does not exist. I think this may fix it:

```
def setUp(self):
    request.environ['wsgi.input'] = BytesIO()
    request.environ['wsgi.input'].write("")
    request.environ['wsgi.input'].seek(0)
```

py4web/utils/form.py
ndegroot: Oops, I’ll check this and change the test. And I will investigate when the decode error occurs. Not always apparently. Which version of jwt are you using?

PR #393, opened 2020-12-26
use single _ on internal hyperlinks and double (anonymous target) on …

paolopas:
…external ones


paolopas:
Internal links should use sphinx ref roles and labels, e.g.
.. _section xyz:
Section xyz
=========
...
:ref:`section xyz`

paolopas:
for the syntax of a reference name see
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#reference-names

PR #407, opened 2021-01-02
Auth - remove debugging print statements

jpsteil:
I inadvertently left some print statements in when debugging.  I Cleaned this up


mdipierro:
Pretty sure this was my fault! thanks for fixing it.

PR #408, opened 2021-01-03
chapter review: make use of ReST comments for things to fix (search f…

paolopas:
1. fixed examples for Python3 (most prints and long ints)
2. addes ReST comments (with FIXMEs) for things still to fix
3. fixed internal crosslinks (maybe @nicozanf is interested in actual usage of :ref:)
4. somewhere needed to cite web2py explicitly

and more


paolopas:
disable syntax highlight on literal blocks (::)

- speed up sphinx
- force authors to use code-block directive (.. code:: ...) with right Pygment lexer alias

paolopas:
see
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rst-literal-blocks

nicozanf:
It's OK for me. Using the FIXME ReST comment is a good idea. Grazie Paolo!

paolopas:
Near 1 MB space saved in docs/images, pdf goes from 3.2 MB to 2.4 MB

PR #413, opened 2021-01-06
Create missing standard folders inside existing apps

nicozanf:
Even if this is not so beautiful, I think it avoids many errors - especially for newbies.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -1039,6 +1039,12 @@ class Reloader:
                     # forget the module
                     del Reloader.MODULES[app_name]
                     clear_modules()
+                # create missing standard folders
+                standard_folders = ["databases", "static", "translations", "templates"]
+                for folder in standard_folders:
+                   folder_path = os.path.join(path, folder)
+                   if not os.path.exists(folder_path):
+                      os.mkdir(folder_path)                
                 module = importlib.machinery.SourceFileLoader(
                     module_name, init
                 ).load_module()


mdipierro:
Let's discuss these. Not all apps require this. This is a convention used by common.py in the default app but I do not think we should always enforce it. Maybe this belongs in common. Maybe we should add a function that creates missing folder and common specifies which ones are used/needed.

jpsteil:
I agree

I don't want those folders created when I do a minimal-type app. I want to
create then when I want to use them.

Jim

agavgavi:
Both the `DAL` and `Translator` class takes in the `settings.___FOLDER` to where it will be stored. Yes these parameters are being passed along to the superclass, but could it be used in core.py to take those folder urls and create them if they don't exist as well?

Not sure if we could do something similar with the `Template` class and `templates` or `static` but that could work for at least 2 of the folders. This way it would be based on whatever folder is specified in `settings`, and if you don't use either, it would never be created.

mdipierro:
can's we just do in common.py

for folder in [...]:
    if not os.path.exists(folder):
       os.mkdir(folder)

nicozanf:
Well, I'm not sure what to think about this...
I've created it for trying to avoid to add a warning on almost every page in the manual, hence adding its logic to common.py does not resolve the problem. But I like the suggestion of @agavgavi to implement in another way, just when it's really necessary (but it's too hard to implement it by myself).
Anyhow I agree that this is not such a big problem and surely in minimal-type apps it's a stretch. Feel free to just close the PR  if it's not so useful.

mdipierro:
Addresses in a different manner in commit 64a91bc7bb7dcac2edab8a61c87f158d6390668e

PR #420, opened 2021-01-13
fix URL(use_appname=False) in shell

### LINKED ISSUE #419: URL(use_appname=False) bug
```
[pax@localhost py4web]$ ./py4web.py shell apps
Python 3.8.6 (default, Sep 25 2020, 00:00:00) 
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from py4web.core import URL
>>> URL(use_appname=False)
  File "/home/pax/.local/lib/python3.8/site-packages/bottle.py", line 1397, in __getattr__
    var = self.environ['bottle.request.ext.%s'%name]
KeyError: 'bottle.request.ext.app_name'

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/home/pax/Job/py4web/github/py4web/py4web/core.py", line 614, in URL
    if (request.app_name != "_default" and use_appname)
  File "/home/pax/.local/lib/python3.8/site-packages/bottle.py", line 1400, in __getattr__
    raise AttributeError('Attribute %r not defined.' % name)
AttributeError: Attribute 'app_name' not defined.
```

paolopas:
fix #419

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -601,7 +601,7 @@ def URL(
     """
     if use_appname is None:
         use_appname = request.headers.get("x-py4web-appname")
-        use_appname = True if use_appname is None else not use_appname
+        use_appname = True
     script_name = (
         request.environ.get("HTTP_X_SCRIPT_NAME", "")
         or request.environ.get("SCRIPT_NAME", "")
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -600,8 +600,7 @@ def URL(
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
     if use_appname is None:
-        use_appname = request.headers.get("x-py4web-appname")
-        use_appname = True if use_appname is None else not use_appname
+        use_appname = request.headers.get("x-py4web-appname", "") == ""
     script_name = (
         request.environ.get("HTTP_X_SCRIPT_NAME", "")
         or request.environ.get("SCRIPT_NAME", "")
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -600,8 +600,7 @@ def URL(
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
     if use_appname is None:
-        use_appname = request.headers.get("x-py4web-appname")
-        use_appname = True
+        use_appname = not request.headers.get("x-py4web-appname")
     script_name = (
         request.environ.get("HTTP_X_SCRIPT_NAME", "")
         or request.environ.get("SCRIPT_NAME", "")
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -611,7 +611,7 @@ def URL(
     else:
         prefix = script_name + (
             "/%s/" % request.app_name
-            if (request.app_name != "_default" and use_appname)
+            if (use_appname and request.app_name != "_default")
             else "/"
         )
     broken_parts = []
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -600,7 +600,8 @@ def URL(
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
     if use_appname is None:
-        use_appname = not request.headers.get("x-py4web-appname")
+        use_appname = request.headers.get("x-py4web-appname")
+        use_appname = True if use_appname is None else not use_appname
     script_name = (
         request.environ.get("HTTP_X_SCRIPT_NAME", "")
         or request.environ.get("SCRIPT_NAME", "")
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -593,33 +593,35 @@ def URL(
 ):
     """
     Examples:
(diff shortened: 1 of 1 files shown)


jpsteil:
I began getting this error this afternoon even when use_appname = True. I tried this fix, but ended up with this on the same line of code to fix:

if (hasattr(request, 'app_name') and request.app_name != "_default" and use_appname)

instead.

It's weird because all was working fine and out of the blue, this started happening.

jpsteil:
Retract that, now my URLs aren't building properly.  Will post more when I get it working.

jpsteil:
Looks like my issue is related to building URLs in a method called by Inject, and not related to this issue.  I will research my issue more and submit an issue if I can't resolve it.

valq7711:
`request.app_name`  is set in action decorator, i.e. URL is intended to use in action

paolopas:
@jpsteil my fix is simpler, request.app_name is never evaluated when use_appname == False,
take a look at diff

jpsteil:
Paolo

You are correct. My errors had to do with my use of URL in an inject and
not use_appname. Please disregard my comments.

paolopas:
@valq7711:
The fact that URL should be used into actions only might be questionable.
Surely the documentation does not make any mention of that and we know that "poor documentation makes code poor".
I think that fixing documentation is a good way to contribute to open source, (developers are lazy guys...), and in fact I found the problem testing some examples of docs/chapter-10.rst.

@mdipierro:
It is docs/chapter-10.rst the right place to state that? Because it is a big change from web2py.

Or maybe we can accomodate to let URL be used by the cli shell too?

Anyway you can change the description of this pull request to whatever you want and the fix still remain valid, it is a matter of efficiency on expression evaluation.

paolopas:
@valq7711 please can you tell which one of the last two commit is the right one?

paolopas:
I think it is time for a break...

valq7711:
@paolopas 
The logic is as follows:
If use_appname is `None`, then it is not passed, in which case it should be set to `True` (default), unless the application is mapped to a domain (`x-py4web-appname` indicates that app is mapped).
So you're right, this:
https://github.com/web2py/py4web/blob/ad764fce99740d35293814221966c1ca7843a82e/py4web/core.py#L602-L604
may be collapsed to this:
```python
if use_appname is None: 
     use_appname = not request.headers.get("x-py4web-appname") 
```

paolopas:
Someone said **Readabilty counts**

paolopas:
Someone said **Readabilty counts**

py4web/core.py
```diff
@@ -600,8 +600,7 @@ def URL(
     URL('a','b',vars=dict(x=1),use_appname=False) -> /{script_name?}/a/b?x=1
     """
     if use_appname is None:
-        use_appname = request.headers.get("x-py4web-appname")
-        use_appname = True if use_appname is None else not use_appname
+        use_appname = request.headers.get("x-py4web-appname", "") == ""
```
valq7711, line 603: Fine!

valq7711:
@paolopas have you seen https://github.com/paolopas/py4web/pull/1/files?

py4web/core.py
paolopas: Your version sounds good
use_appname = not request.headers.get("x-py4web-appname")

jpsteil:
So, will this fix allow you to use URL in Inject?

valq7711:
@jpsteil  Jim, show your snippet please, as I can't figure out the reason why it doesn't work

valq7711:
@jpsteil  maybe you do something like `foo = URL()` in inject?

jpsteil:
Val - I'll pull in the PR and give it a test later today.

-Jim

paolopas:
has_appname added for clarity

paolopas:
I hope that is not necessary to explain that in Python `A and B`
does not necessarily evaluates B (as for `A or B`)

paolopas:
@valq7711 for me there is nothing more to do, do you still have some doubts?

valq7711:
@paolopas  perfection is unattainable ;)
https://github.com/paolopas/py4web/pull/2/files

mdipierro:
I think URL can be used in actions and functions/methods called by actions. This should be stated where URL is first used in the docs and in chapter 10. If URL is used in a place where app_name is not passed and undefined it should raise an error (with assert).

paolopas:
A good place could be the function docstring too

paolopas:
@mdipierro
I think there are many reason for URL to do not enforce app_name (as it does now):
1. be tolerant if there is no strict reason to be not make code more robust,
1. actually when we use URL in _default app, it behave as if no app_name
   is present, so why do not do the same in cases app_name
   is really missing? (for whatever reason)
1. maybe an action could be called in py4web shell environment, without
   the need to give py4web shell a virtual app_name (e.g. py4web-shell)

Otherwise the right place to document should be chapter 15 ("From web2py to py4web")
since enforcing app_name is a big change with respect to web2py.

Issue #411, opened 2021-01-04
Upload Fields and PR #379


agavgavi:
_Copied from the py4web group for extra visibility linked [here](https://groups.google.com/g/py4web/c/5hl6YEANUmI)_

Good afternoon,

I've been working with an application that deals with file uploads through the Form class. After PR #379's change, uploads will no longer occur in my program due to the upload field not showing up in bottle's request.forms.

By doing ` if not field.name in post_vars: continue` since upload fields aren't stored in the request.forms FormDict object and instead [stored in request.files](https://bottlepy.org/docs/dev/tutorial.html#file-uploads), py4web doesn't know how to handle file uploads.

Am I doing something wrong? My field is defined as:
```python
Field(
       "upload_file",
        "upload",
        label=T("Dataset Upload (CSV)"),
        uploadfolder=UPLOAD_FOLDER,
    )
```

and before I could just access the filename through `form.vars['upload_file']`, but now the entire processing loop is skipped for upload fields in the Form class, so it will never appear.

Thank you for any suggestions anyone has on fixing this, for now I've just commented out the if statement that was added in the PR and things return to how it was before.

valq7711:
I think `request.forms` should be replaced with `request.POST`. Can you try/test?

valq7711:
`request.POST = request.forms + request.files`
that's the only difference

agavgavi:
Worked perfectly, thanks for the info, I can make a PR or you can if you want.

valq7711:
it's better if you do it ;)

сб, 16 янв. 2021 г. в 02:44, Andrew Gavgavian <notifications@github.com>:

agavgavi:
Very true, I've fallen back from being in the top 5 contributors last spring, got to up my game to be back in the leagues of the big shots like you guys, haha :P

Kkeller83:
Is this a pr for the py4web Form.py or how is it related?

You seem to be using bottle specific Form classes with pydal no?

Anyhow maybe you guys can elaborate a bit more what's going on here please?

Thanks!

valq7711:
@Kkeller83 
bottle `request.forms` and `request.files` are just derived properties from `request.POST` (they are just result of filtering of `request.POST` and nothing else)
Using `request.forms` in `form.py` forces you to hold file upload yourself (by request.files) like in this example
https://groups.google.com/g/py4web/c/2ZYc5yHMgDQ

This condition is always `False` as there are no files in `request.forms` by design
[...]

PR #421, opened 2021-01-16
Updated form to re-allow file uploads, thanks Val!

agavgavi:
This change will fix issue #411 and this [post ](https://groups.google.com/g/py4web/c/2ZYc5yHMgDQ)on the google group where `request.forms` doesn't keep track of uploaded files. Thanks Val for the insight!


mdipierro:
We should change the docs to mention request.POST in place of request.forms everywhere

agavgavi:
I'll go through and make a PR for this documentation change soon

PR #422, opened 2021-01-17
xapi introduction

valq7711:


valq7711:
Now we have extremely cool api factory ;) 
(Flask blueprints smoke quietly on the sidelines)

valq7711:
```python
# ---------------- [ Experimental API for general purposes ] ----------------------
## Usage example

from py4web.utils.xapi import API
from py4web import action
import datetime

api  = API.factory()  # create api wrapper
api.add_proxy('uses') # declare decorator with deferred resolution

# create filters/validators
P = api.make_filter('path')
Q = api.make_filter('query')
B = api.make_filter('body')

@api()
class MyAPI:
    # declare/init all thread safe attrs in __init__
    @api.thread_safe(safeguard = True)
    def __init__(self, foo):
        self.now = None
        self.foo = foo
        self.method = None

    # @api.uses(T)  # `global` fixtures
    def on_request(self, route_ctx):
        self.method = route_ctx['route']['method']
        self.now = datetime.datetime.now()

    @api.get('/')
    @api.get('<id>')
    @api.uses(db)
    def todos_list(self, id = P.id.int|None):
        if id:
            return db(db.todo.id == id).select().first().as_dict()
        else:
            return dict(items=db(db.todo).select(orderby=~db.todo.id).as_list())

    @api.post('/')
    @api.uses(db)
    def create(self, info = B.info):
        return dict(id=db.todo.insert(info=info))

    @api.put('<id>')
    @api.uses(db)
    def update(self, id = P.id.int, info = B.info):
        return dict(id = db(db.todo.id == id).update(info=info))

    @api.shaper()  # like `Fixture.transform`
    def shaper(self, ret):
        ret['timestamp'] = self.now
        return ret

    @api.on_success()
    @api.on_error(P.ValueError)  # @api.on_error()  - catch all errors
    def on(self, err = None):
        if err:
            raise HTTP(f'400 something went wrong: {str(err)}')


# api context - adapter between framework and api:
# context-object must have:
#   .get(ctx_property:str) -> callable -> ctx_property:dict
#   .__call__(**path_params)
#   .__contains__(ctx_property:str)
# ctx_properties should match filters/validators, i.e.:
#   if we have `P = api.make_filter('path')`
#   then `ctx.get('path')` must return callable that returns ctx_property

class API_CTX:
    def __init__(self):
        self.ctx = dict(
            path = lambda: None,
            query = lambda: request.query,
            body = lambda: request.json,
        )
    def __call__(self, **args):
        self.ctx['path'] = lambda: args

    def get(self, k):
        return self.ctx.get(k)

    def __contains__(self, p):
        return p in self.ctx

# set mounter and resolve deferred decorators
api.proxy(mounter = action, uses = action.uses)

# set api context
api.set_ctx(API_CTX())

# will be mounted at `app/xtodo`
MyAPI.premount('xtodo')

# run api
MyAPI.run(foo = 'foo')

# instead of last 2 above we can
# MyAPI.mount('xtodo', args = ['foo'])
# or MyAPI.mount('xtodo', kwargs = dict(foo = 'foo'))
```

mdipierro:
Thank you Val. You provide an example of usage of all the APIs. Can you provide a simpler/minimalist example? Can you also provide a rationale? What can we do with this that cannot be done with bare-bone py4web? Or how does it make things simpler?

valq7711:
minimalist example ))
```python
from py4web import action, request 
from py4web.utils.xapi import Param, make_decorator

class API_CTX(dict):
    # just for using as enum - see `Q = Param(...)` below
    class Props:  
        query = "query"
    def __init__(self):
        self[self.Props.query] = lambda: request.query
    def __call__(self, **args):
        pass

api = make_decorator(API_CTX())
Q = Param(API_CTX.Props.query)

# without this we will get 500 error (off course it is temporary workaround)
def catch_param_error(f):
    @functools.wraps(f)
    def catcher(*args, **kw):
        try:
            ret = f(*args, **kw)
        except Param.ValueError as err:
            raise bottle.HTTPError(400, body = str(err))
        return ret
    return catcher


# all above could be placed in common.py
# and then just `from common import  api, Q`

@action('echo_query')
@catch_param_error
@api()
def echo_query(offset = Q.skip.int|0, limit = Q.limit.int):
    return dict(offset = offset, limit = limit)

# try:
# should be OK
# app/echo_query?skip=10&limit=100 
# app/echo_query?limit=100 

# should raise 400
# app/echo_query?limit=blah 
# app/echo_query?skip=10
# app/echo_query?limit=100&whatever=45
```

mdipierro:
What is the reason for all of the API_CTX builerplate? Why not exposed only
api from the module and make catch_param error an attribute of api?

valq7711:
@mdipierro yes of course there are many points to polish, what about dev-branch?

mdipierro:
sounds good

valq7711:
@mdipierro 
I added `apps/xtodo` for demo
## Rationale
- `some/path/<id:int>` returns not found if `id` is not `int` , but in case of api  it is better to return `bad request`
- modular approach (api-module can be written with import only xapi and nothing else. xapi itself is framework agnostic)
- `uses()` applies all fixtures unconditionally


```python
# --------------- [ traditional approach] ---------------
@action("api/<id:int>", method="DELETE")
@action.uses(session)
@action.requires(user_in(session))
@action.uses(db)
def todo(id):
    db(db.todo.id == id).delete()
    return dict()

# --------------- [ xapi approach ] ---------------
...
P = api.make_filter('path', allow_extra = False)
F = api.make_filter('fixtures')

@api()
class todo_api:
...
    @api.delete('<id>')
    def delete(self,
        id = P.id.int,
        user = F.session['user'],   #  fixture.on_request() is invoked only when/if you touch it
        db = F.db                   #  db.on_request() won't be called if 'user' is not in session
    ):
        db(db.todo.id == id).delete()
        return dict()

```

valq7711:
I'm git-confused

PR #423, opened 2021-01-19
Updated mtable for additional features

agavgavi:
Mtable in the dashboard works great for viewing tables, but a few attributes don't display in there.

The table assumes that rational numbers are displayed if the `field.type == 'float'` but float doesn't seem to be a type in pydal going off the documentation online, instead the field can either be double or decimal.

In addition it doesn't support displaying the pydal `list:<>` type. This current version has `list:integer` integrated with it, but it wouldn't be too terrible to add the others as well in a future commit. Note that this will display the value as a comma-delineated list in brackets, and to update it you have to keep the same syntax, but it does allow the field to even appear in the table.

### DIFF
--- a/apps/_dashboard/static/components/mtable.html
+++ b/apps/_dashboard/static/components/mtable.html
@@ -32,13 +32,15 @@
             <span v-else><i class="fa fa-square"></i> None</span>
           </span>
           <input class="input" v-else-if="field.type=='integer'" type="number" step="1" v-model="item[field.name]" v-bind:readonly="!editable">
-          <input class="input" v-else-if="field.type=='double'" type="number" step="0.01" v-model="item[field.name]" v-bind:readonly="!editable">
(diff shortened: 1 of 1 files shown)


agavgavi:
Re reading through sample code, it does seem like `float` is a type in pydal, it's used in some tables in the examples app, but in the [Field Types](http://py4web.com/_documentation/static/en/chapter-05.html#field-constructor) section of the Documentation in both the py4web docs and web2py, the `float` type doesn't seem to exist.

Is this missing from documentation somewhere? Or are the two names analogous?

agavgavi:
After adding `list:string` support and giving `float/decimal/double` proper increment steps, it seems that there is an issue with the pydal rest-api when interacting with lists. No matter the format, even if a list isn't displayed on the mtable, when saving the fields changes it will error out due to the rest-api not liking the values of the list. 

I'm not sure what the proper serialization for that would be, and I'm going to move this part over to an issue, but in terms of viewing the fields in the database, I think this is still a great addition to be able to see list fields.

PR #426, opened 2021-01-20
adding test for field.uploadfolder to form.py

ali96343:


py4web/utils/form.py
```diff
@@ -474,7 +474,7 @@ def __init__(
                             if field.type == "upload":
                                 value = request.files.get(field.name)
                                 delete = post_vars.get("_delete_" + field.name)
-                                if value is not None and hasattr(value, "file"):
+                                if value is not None and hasattr(value, "file") and hasattr(field,'uploadfolder') and field.uploadfolder:
```
valq7711, line 477: - I suppose if `value is not None` it must have `file`-attr, no?
- `upload` field always has `uploadfolder`-attr, therefore hasattr(field,'uploadfolder') is redundant

Also, your patch has side effect, since there is `elif self.record and not delete:`  that will be processed even if value is not `None`  (if `uploadfolder` is not set)

I think it should be:
```python
if value is not None:
    if field.uploadfolder:
        value = field.store(
            value.file, value.filename, field.uploadfolder
        )
elif self.record and not delete:
    ...
```
just in case: `hasattr()` is not the same as `key in somedict`, `hasattr()` is just a shortcut for:
```python
try:
   getattr(...)
except AttributeError:
    return False
return True
```
so if you need attr-value it is better (for performance) to use `attr_value = getattr(obj, 'attr', None)` 
`hasattr()` is only suitable for logic like:
```python
if not hasattr(obj, 'foo'): obj.foo = ... 
```

Issue #424, opened 2021-01-19
clarify HTTP exceptions handle logic


valq7711:
So we have `uses` wrapper and `action` wrapper (see below).
The question is, how can I raise an **HTTP** exception so that the fixtures consider it an error + transparently pass something useful to api-client and keep (or not) cookies/headers?
It might be better to have an HTTPError exception (like bottle has) with some parameters, because I'm totally confused.

https://github.com/web2py/py4web/blob/a8fa1d02436250a57b5619f48e0e7e0ea369a3d7/py4web/core.py#L725-L737
https://github.com/web2py/py4web/blob/a8fa1d02436250a57b5619f48e0e7e0ea369a3d7/py4web/core.py#L765-L789

PR #428, opened 2021-01-21
add type-attr to HTTP

valq7711:
allow rise HTTP with body and `on_error` - triggering (`if http.type != 'success'`)


valq7711:
related to #424

valq7711:
I think we should change Fixture `on_error(self)` to `on_error(self, status=None)`
Sooner or later it will become an issue.

PR #429, opened 2021-01-21
xapi introduction

valq7711:
reborn #422


mdipierro:
Although this is very well done I find the syntax complex and I do not see the benefit. It is very well possible I am missing something.

In your example https://github.com/web2py/py4web/pull/422#issuecomment-764158600 it seems to make the code more verbose than the py4web barebone case.

You can already reduce boiler plate by defining:

def api(uri, method='GET', fixtures=[session, db, auth.user]):
     return lambda func: action(uri, method=method)(action.uses(*fixtures)(func))

and then do

```
@api('todo/<id:int>', 'DELETE')
def todo(id):
    db(db.todo.id == id).delete()
    return dict()
```

valq7711:
Of course, xapi is just an alternative approach to avoid using globals - there is a small source of bugs - if you don't include a fixture in action.uses (...), the action might still work, but with unexpected behavior.
There is another advantage - since xapi mounts routes in api.mount() -call, you can just import and mount it in multiple applications.
To achieve the same with py4web barebones, you need to wrap all `@action` in a function (as in auth.py).
So it makes it easier to develop modules for distribution.
As for me explicite declaration in arguments like  `offset = Q.skip` is more clear, then `offset = request.query.skip` somehere in the function body + it allows to generate api documentation

PR #430, opened 2021-01-22
new test for field.uploadfolder to form.py

ali96343:
test with Valq remarks


mdipierro:
I am missing the context here

ali96343:
without this patch it is not possible to use the 'upload' field
without the uploadfolder field

for example
it doesn't work:   Field( 'xls_file', 'upload', ),
it works :  Field( 'xls_file', 'upload', uploadfolder=' some path' ),

But, my application doesn't need to save the file, I only need to
upload the file to the controller

PR #431, opened 2021-01-24
patch cli+allow use gevent+websocket and wsgirefThreading

valq7711:
new cli-option -s/--server:
```python
Options:
   ...
  -s, --server [default|wsgiref|tornado|gunicorn|gevent]
                                  server to use  [default: default]
   ...
```
run using gevent + monkey.patch_all
```python
py4web usegevent run -s gevent
```


valq7711:
now you can:
```
py4web.py usegevent run -s geventWebSocketServer
```

valq7711:
now you can
```
py4web.py run -s wsgirefThreadingServer
```

valq7711:
@mdipierro 
I think we can now use `wsgirefThreadingServer` as  windows default server

mdipierro:
How about using rocket from web2py instead. I can make it its own module on pypi. Given one more day to review this.

valq7711:
There is an issue with rocket on windows, I had similar problem with files more than 32 mb on other browsers (switching to web2py_gevent.exe solved the problem)
https://github.com/explorigin/Rocket/issues/1

valq7711:
waitress works fine on windows, just don't mix it with gevent

valq7711:
@mdipierro I think it is ready to commit!

mdipierro:
There are lots of caveats with gevent including the fact that also psycopg2 and sqlite require monkey patching.

PR #441, opened 2021-02-07
Fix server option + add sphinx-tabs extension

nicozanf:
Currently, if you run py4web without any option you get an error instead of the help page.
Also, the sphinx-tabs extension will be used later on the manual


valq7711:
@nicozanf     My bad ( I got that it could happen after merge;) thank you!

PR #449, opened 2021-02-14
fixed apps_folder usage

paolopas:
Unchecked (i.e. not existing) apps_folder allowed on **setup** and **new_app** commands only.
Other py4web commands need existing apps_folder and ask for creation of missing `__init__.py`.
See #444 for other fixed issues.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -9,7 +9,7 @@ import code
 import copy
 import datetime
 import functools
-import importlib
+import importlib.util
 import importlib.machinery
 import inspect
 import json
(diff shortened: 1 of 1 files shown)


mdipierro:
Let's talk about this. The apps folder name could have any arbitrary name and it must be specified when py4web starts. BUT when an app wants to include a module from another app it should just be able to do: "import apps.otherapp.modulename". In other words I want the module that corresponds to the folder to always be referred to as "apps" in code. I think this PR break that

mdipierro:
I think unchecked app folder should be allowed for setup only. I made a commit for that following your suggesting to address failing tests.

paolopas:
Massimo,
I was complaining about harcoded 'apps' since my PR #438 ...

well, in that case I suppose that you want to see my wonderful

paolopas:
More important, now "import apps.sub" does not relies on having the APPS_FOLDER visible from current working directory to work!

PR #455, opened 2021-02-19
adding socketio, websocket to aiohttp, tornado, threading, gevent

ali96343:
new servers list
'wsgirefAioSioWsServer', 'tornadoSioWsServer', 'geventSioWsServer', 'wsgirefSioServer',


valq7711:
@ali96343 
what about this? https://github.com/etianen/aiohttp-wsgi/issues/11
and especially this?  https://github.com/etianen/aiohttp-wsgi/issues/7

Gevent-websocket server is thread-safe, so you can directly use thread-base things like pydal in websocket request. But with these servers everything that is run outside of wsgi-handler (i.e. async-features like websockets) is not thread-safe, is it?

Cold shower, for those who consider async a panacea  https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/

ali96343:
"run outside of wsgi-handler"
The interaction between the websocket server and the wcgi-pydal-app can be performed via redis and 
http_client_fetch (wsgi-app-controller-with-pydal) .This is the standard solution from flask-django-world
( of course async is not a panacea and python is not erlang-elixir ) 
aiohttp is supported by the core python developers, bugs are fixed

mdipierro:
Give me some time to digest this. :-)

mdipierro:
I like your addition to anyserver,py although I think it should go in utils and be called wsservers.py. I do not see where that is imported i your examples. Also you have two example apps. why? Could you provide some documentation for the serverside usage?

ali96343:
Today it is possible to use two protocols (socket.io and websockets) to build a chat server.
The only place where we can add these protocol handlers is the py4web file anyservers.py. The protocol handlers (socket.io and websockets) are launched by starting the server from the file anyservers.py .

For the websockets handler native server handlers are used. For the socketio handlers, we use the nice library by Miguel Grinberg - python-socketio.

Unfortunately, I don't know how to put anyservers.py  (wsservers.py) in the utils directory with the current py4web architecture.

To test the two installed protocols I use two test applications:  _socketio and _ws.

The import libraries are done in the app-index.html files.
in _socketio-index.html  "<script src="https://cdn.socket.io/3.1.1/socket.io.min.js"></script>"
in _ws-index.html " ws = new WebSocket('ws://127.0.0.1:8000/'); "

Also at this link https://github.com/ali96343/py4web-chat , there is a more realistic chat server and the file anyservers.py

ali96343:
I figured out how to put anyservers.py to utils/siowsservers.py
sorry
I'll rewrite the patch.

PR #458, opened 2021-02-22
Update chapter-06.rst

dantoac:
update url to owasp SameSite article


dantoac:
just a url link update to the SameSite article in owasp web.

PR #463, opened 2021-03-01
Removed jwt from URLSigner in favor of the simpler hashlib

lucadealfaro:
I removed the use of jwt from the URLSigner in favor of the simpler hashlib.  This enables also more careful control of what goes into the signature, including the salt.  The main motivation of the PR is to avoid breakages due to jwt incompatibilities before and after version 2.0.0.


mdipierro:
Maybe this is for the best

PR #464, opened 2021-03-02
onclick javascript action for grid custom buttons

Kkeller83:


Kkeller83:
Looks like something went wrong with this PR.. I only changed 3 lines and it shows way too man lines changed.. I will try again. This can be deleted.

PR #475, opened 2021-03-18
Validation of Uploads in Forms

agavgavi:
Currently even if the form's validation fires and a file upload doesn't pass the `field.requires` field, it will still upload and store the file. 

Personally I believe that the file should only be uploaded _if_ the file passes all validations.

This is also the same with downloading the file, we would only want that if the file successfully uploaded.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -181,7 +181,7 @@ class FormStyleFactory:
                 )
             elif field.type == "upload":
                 control = DIV()
-                if value and not error:
+                if value and field.download_url is not None and not error:
                     download_div = DIV()
                     download_div.append(
                         LABEL(
@@ -475,6 +475,7 @@ class Form(object):
                 record_id = self.record and self.record.get("id")
                 if not post_vars.get("_delete"):
                     validated_vars = {}
+                    uploaded_files = []
                     for field in self.table:
                         if field.writable and field.type != "id":
                             original_value = post_vars.getall(field.name)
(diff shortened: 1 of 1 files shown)


valq7711:
There is one more issue - what if upload field will be followed by another error-field

agavgavi:
So you're saying that if the upload passes validation, but a later field in the form doesn't it would still upload. That is correct, maybe we could move the `field.store()` call on line 500 to be after all validation? I'm not sure the best way to about that if there are multiple file uploads though.

valq7711:
I think we can collect upload fields in a separate list and perform `store` after and only if whole form  is valid

valq7711:
The best way to perform validation is doing it via ajax in this case we can keep file-fields values

mdipierro:
> I think we can collect upload fields in a separate list and perform `store` after and only if whole form is valid

Yes. I think we should do that. Would you be able to submit a PR for that or amend this PR?

agavgavi:
I can accomplish this and amend this PR for that.

mdipierro:
Fantastic. Thank you!

agavgavi:
Whenever a user uploads a file it is appended in the `uploaded_files` list. After **_all_**  validation has occurred we loop through all files in uploaded_files and call a field.store on them. And then if it's successful we will adjust the value in `validated_vars` since pydal will change the name of the file uploaded.

mdipierro:
nice. thank you.

PR #485, opened 2021-03-31
Update chapter-07.html

JMBeresford:
various typo and omission fixes


JMBeresford:
wrong location for PR

PR #495, opened 2021-04-16
Fixed form for select fields for bulma

lucadealfaro:
The formatting for select fields in Bulma was broken in FormStyleBulma, and unfortunately, to fix it, I had to add a new parameter to the FormStyle factory.  I hope this is ok... another solution would be to write FormStyleBulma completely independently from the normal FormStyle, to better handle Bulma idiosyncrasies, but given that everything almost worked, I opted for this more local fix.  Let me know what you think.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -62,11 +62,14 @@ class FormStyleFactory:
             "select": "",
             "textarea": "",
         }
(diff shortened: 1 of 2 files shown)


mdipierro:
Do not like the name class_inner_exceptions but I see the problem. :-)

PR #500, opened 2021-04-23
Fixing the missing folder "databases" and "sql.log" file in the myfeed app.

Kkeller83:
The missing folder and file causes a bunch of error messages on startup


agavgavi:
We probably shouldn't push the actual database to the repo since that could contain private info. We just need the folder to be known which can be done either by putting a `README.md` with no contents in it in the folder, or a `.gitkeep`.

This error pops up a lot because every so often the README is deleted and the folder disappears.

Kkeller83:
Well this SQL.log is also just an empty file. But technically you are right.

mdipierro:
Yes. It should not be committed

Kkeller83:
Maybe you have a moment to put a fix up. If anyone wants to try py4web now the error messages are quite intense.

Will close this.

Issue #501, opened 2021-04-25
Unauthorized Application/Json Request Not Redirecting


macneiln:
In auth.py the `abort_or_redirect( )` function is raising a 403 Forbidden error when the request `accept` header contains `application/json` instead of doing the standard redirect/appending the next url as it does when the page is loaded directly (not via axois, etc.)

This doesn't seem to add any value that I'm aware of but instead complicates `application/json` requests where the user is not authorized, logged in, etc. and client-side my application is not redirected as anticipated.

My proposal is either to remove it so it functions the same regardless of request type, or if this functionality is required for something then we should add a secondary header to determine whether this functionality should be bypassed if required.

### auth.py
```
    def abort_or_redirect(self, page, message=""):

        """
        return HTTP 403 if 'application/json' in HTTP_ACCEPT
        else redirects to page
        """
        if re.search(REGEX_APPJSON, request.headers.get("accept", "")):
            abort(403)

        redirect_next = request.fullpath
        if request.query_string:
            redirect_next = redirect_next + "?{}".format(request.query_string)
        self.auth.flash.set(message)
        redirect(
            URL(
                self.auth.route,
                page,
                vars=dict(next=redirect_next),
                use_appname=self.auth.param.use_appname_in_redirects,
            )
        )
```

valq7711:
can you provide an example of how you hold redirect using axios?

valq7711:
I mean what does your client side code look like?

macneiln:
> can you provide an example of how you hold redirect using axios?

- I use vuejs, vue-router and axios client-side. 
- Axios sends the request for the page the user has clicked on and the browser automatically handles redirects if they occur and provides me the content from the secondary request (if a HTTP 303 was received), so I end up with the data I'm looking for either way. In this case it would redirect the user to login page before they can get to the page for the link they clicked.

valq7711:
in case of ajax and server-side redirect you will just get full rendered html of login page in the response, so what do you do next? Go to redirected url with full reloading?

macneiln:
I have a custom fixture that handles the different types of requests I send and in the case of the auth specific routes I override/serialize the auth form in my fixture transform function so that it can be returned via json rather than the default auth.html page so no full reload is required once initially loaded.

valq7711:
So it should be 
>This doesn't seem to add any value that I'm aware of  **(as I have custom fixtures that override/serialize auth forms in appropriate format)**

;)

PR #502, opened 2021-04-25
Application/json request updates

macneiln:
Adding additional header to allow use of standard redirection when request is application/json and the use is unauthorized, not logged in, etc.


macneiln:
Resolves #501.

valq7711:
I think it is better to allow `abort_or_redirect` method to be overridden than to introduce a framework specific header.

valq7711:
Or something like `on_abort_or_redirect` callback as Auth param

PR #505, opened 2021-04-27
adding socketio, websocket to aiohttp, tornado, threading, gevent - u…

ali96343:
These changes are in response to PR comments
- moved anyservers.py  to utils/wsservers.py
- renamed anyservers_list to wsservers_list
- added  usage-comments  to _ws-app, _socketio-app, wsservers.py


ali96343:
probably not all users need web sockets and extra libraries.
so wsservers.py is renamed to wsservers.py.txt

mdipierro:
I will merge this PR but I will then move the examples apps into examples/

mdipierro:
I moved the examples under apps/examples. Can you please check I did not break anything? 
TODO:

- I think there should be an easier way to start the websocket server using a py4web command line option. 
- The websocket should always have a /appname/{channel} path since different apps should not mix up they websockets.
- An app should have a way to pass an authentication function to its own websocket to restrict who can connect and who can post.

PR #506, opened 2021-04-27
Grid - formatters, htmx, bulma class/style

jpsteil:
I understand these changes may be controversial and am looking for comments/direction.

Add row to grid formatters. Instead of passing just the value for the specific field, pass the entire row contents.  This allows for more flexibility when formatting grid fields. It feels a little hackish the way I implemented the format_field function. Is there a better way to do this? The challenge is to stay backward compatible but to add an optional 'row' parm on the formatter call so you can use the whole row to format your field.  

For example - if I have a row with a customer name, city, state and zip, instead of putting them all in different columns I can now format the name to show all 4.

```
grid.formatters["customer.name"] = lambda value, row: (
    f"{row.id} - {row.name}\n{row.city}, {row.state} {row.zip_code}"
)
```

now I can have all those fields show up nicely formatted in one cell in the grid.

htmx - finish out grid support for executing within an htmx element - sorting, paging, deleting

This PR also updates some of the class/style attributes for a grid using bulma formatting
- make tables responsive
- set numbers to default to right justified
- all grid cells are vertically aligned to the top of the cell

fixed typo on innerHTML


py4web/utils/grid.py
```diff
@@ -1159,3 +1192,55 @@ def parse_referer(r):
...
+        parent_id = parent_id.split("&")[0]
+    except:
+        pass
+
+    return parent_id
+
+
+def format_field(formatter, value, row):
+    try:
+        return formatter(value, row)
```
mdipierro, line 1244: you can do 
```
if formatter.__code__.co_argcount == 1:
```
instead of try except.

py4web/utils/grid.py
```diff
@@ -780,14 +788,33 @@ def render_table_header(self):
...
                 else:
                     sort_query_parms["orderby"] = key
-                    href = URL(self.endpoint, vars=sort_query_parms)
-                    col = A(heading, dw if "~" + key == sort_order else "", _href=href)
+                    url = URL(self.endpoint, vars=sort_query_parms)
+                    if self.param.htmx_target:
+                        rf = parse_referer(request)
+                        if rf and rf.query:
+                            url += "?%s" % rf.query
+                        attr["_hx-get"] = url
```
mdipierro, line 814: Unsure about making py4web dependent on htmx. Could we maybe come up with a grid plugin system to configure this?

jpsteil:
Thanks Massimo. Will look into your suggestions tomorrow.

I have no experience building a plugin system. Any pointers or code/project
examples would be appreciated.


Have you had any time to look at utils.js options vs htmx?

Jim

Kkeller83:
I did some research and from what I understand in order to replicate just the feature to load, replace and pin down some html at a specific spot we would need to use some Ajax (e. g. Fetch etc.) to pull the html code and then replace the div with the html code but using the html tag <embed>.

I don't know how htmx does it.

However i am not sure how a plugin system would really help. It gives the illusion of separation but really if you don't want to use htmx you don't need to use the htmx param.

Meanwhile when you have a plugin system, who maintains the plugins? They cant be maintained separately from grid or py4web anyway.

The alternative plugins avaliable are also pretty slim.
There is alpine.js and stimulus which operate very different from htmx to a point that I wonder how a plugin could help as there ja much more code required, while in htmx its just basically tags.

Unless you literally reimplement htmx style tags with another js alternative or pure js a plugin system just adds overhead and it ends up being maintained by the same people anyway, plus I don't see anyone with the ambition to reimplement htmx.

If you don't use the htmx param you get the grid without htmx targets and then you just use your js solution of choice as usual.

If we come up with our own htmx replacement we would have to maintain that.

And if we have a plugin system with own htmx alternative and htmx with have to maintain a plugin system 2 plugins and our own htmx like functions.

If you really want some sort of code separation maybe it's easier to just have a grid.py and a htmxgrid.py that has the htmx param and targets while the other files doesn't?

mdipierro:
My problem is just with this lines:
```
                    if self.param.htmx_target:
                        rf = parse_referer(request)
                        if rf and rf.query:
                            url += "?%s" % rf.query
                        attr["_hx-get"] = url
```
which appear twice. can we instead pass a function grid.param.postprocessor_plugin = f
where:
```
def f(attr):
      rf = parse_referer(request)
      if rf and rf.query:
           url += "?%s" % rf.query
      attr["_hx-get"] = url
```

the function would live at the application layer and py4web would not need to know how you want to postprocess your attributes.

mdipierro:
I am sure you can propose a better name. Maybe "plugin" is not the correct word.

mdipierro:
I merged this but I am going to make a few changes as suggested above.

jpsteil:
Massimo - I have lots of changes made that I haven't pushed yet. I can
submit a PR with the chances this afternoon if your like.

Jim

mdipierro:
I made the change and now you can do:

```
grid.attributes_plugin['form'] = lambda attrs: attrs.update({
     "_hx-post": request.url,
     "_hx-target": htmx_target,
     "_hx-swap": "innerHTML",
})
```
and similarly for:
```
grid.attributes_plugin['form']
grid.attributes_plugin['link']
grid.attributes_plugin['search_form']
grid.attributes_plugin['button_sort_up']
grid.attributes_plugin['button_sort_down']
grid.attributes_plugin['button_delete']
grid.attributes_plugin['button_page_number']
```
where a plugin a function that takes the current dict of attrs and changes it as it likes.

mdipierro:
Yes. I can look later today

PR #507, opened 2021-04-28
GitHub dashboards

ali96343:
It is suggested to add multistars github dashboards to py4web/apps


mdipierro:
I have difficulty seeing the diff for this PR. I changed 3000 files. why?
Also It contains some files that were already added in a previous PR but moved to apps/examples/ws and apps/example/socketio

agavgavi:
I feel like this would be useful to have in the [py4web showcase repo](https://github.com/web2py/py4web-showcase), but I don't really see the point of adding over 20 different apps to the main py4web application. If the point of py4web is it's speed and small footprint, adding all of these apps by default would be counterproductive to this.

ali96343:
- "difficulty seeing the diff "
 it's probably because the  apps/_ws and  apps/_socketio files have been moved
also the size of the facep4w directory is about 1 Gb 

 - "py4web showcase repo"
 yes, you are right.
We probably need to have a set of links in the documentation to interesting apps.
 Or small app in apps directory with  set of links to interesting apps.
 Maybe I need to reduce the number of  github-dashboards

mdipierro:
I do not think we should commit a 1gb folder. :-)

Of there is an app which is more of a few lines example it belongs it
another repo

ali96343:
ok

agavgavi:
What I see from this is that the py4web-showcase repo should get some attention, either in the readme here, or linked in the documentation. There are some few month old PRs over there for projects, so I think if they were listed and we got more, we could have large cool projects like this still associated with py4web without clogging up this repo.

PR #511, opened 2021-05-03
Added radio button widget and gave no.css some radio button love

Kkeller83:
Added the radio button widget as an alternative for dropdowns in the html forms.py. 
If a field has a "required" validation set the db.field.widget="radio" option can be set to display fields as radio buttons. 
no.css has received some spacing for radio buttons so they look neat.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -166,4 +166,4 @@ Many thanks to everybody, and especially:
 - `Pirsch <https://github.com/Pirsch>`__
 - `sugizo <https://github.com/sugizo>`__
 - `valq7711 <https://github.com/valq7711>`__
-- `Kkeller83 <https://github.com/Kkeller83>`__
+- `Kevin Keller <https://github.com/Kkeller83>`__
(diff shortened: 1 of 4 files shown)


Kkeller83:
Massimo: There are two problems. 1) field.widget is not thread safe 2) if we want to reintroduce it, sure we want to refer to widgets by name and not objects like in  web2py? 

I think we should move away from field widgets and consider widgets a property of forms. For example

form.widgets[field name] = form.default.widgets.boolean 

Where boolean is an object. This is also because forms may one day contain widgets that do not map into fields.

Kkeller83:
@jsteil @lucadealfaro @mdipierro @valq7711 I checked the FromStyle code in forms as I think its a similar approach, so I could asses if I can implement this myself, but I think I am at a loss from my python skills. Could you give me a leg up with this? I would need it to create the poll application for Europython.

Kkeller83:
Will close this. Jim found a new solution. Will create a new pr

PR #512, opened 2021-05-03
Candidate for Attribute plugin

jpsteil:
Here is my proposal for an Attribute Plugin.  

In the previous version I'm not able to send a different URL based on where it is used in the grid.

I've added a class, AttributesPlugin that gives you all the default attributes and is used unless you override it.
```
class AttributesPlugin:
    def __init__(self, target_element=None):
        self.target_element = target_element
        self.default_attrs = {}

    def form(self, url):
        attrs = copy.copy(self.default_attrs)
        attrs["_method"] = "GET"
        attrs["_action"] = url
        return attrs

    def link(self, url):
        attrs = copy.copy(self.default_attrs)
        attrs["_href"] = url
        return attrs

    def confirm(self, message):
        attrs = copy.copy(self.default_attrs)
        attrs["_onclick"] = "if(!confirm('%s')) return false;" % message
        return attrs
```

I've also provided the AttributesPluginHtmx.  I don't know if that should remain in grid.py or if it should be part of the application code.
```
class AttributesPluginHtmx(AttributesPlugin):
    def __init__(self, target_element):
        super().__init__(target_element)
        self.default_attrs = {
            "_hx-target": self.target_element,
            "_hx-swap": "innerHTML",
        }

    def form(self, url):
        attrs = copy.copy(self.default_attrs)
        attrs["_hx-post"] = url
        return attrs

    def link(self, url):
        attrs = copy.copy(self.default_attrs)
        attrs["_hx-get"] = url
        return attrs

    def confirm(self, message):
        attrs = copy.copy(self.default_attrs)
        attrs["_hx-confirm"] = message
        return attrs
```

This fulfills the needs for the htmx implementation as well as the default implementation. I see we could easily build for other technologies or leave that to the individual developers.

For my needs, I didn't need the other element types, search_form, button_sort_up, button_sort_down, button_delete, button_page_number. But we could add methods for those as well, I see where they could be useful for further customization.

I would appreciate any input.


Kkeller83:
COuld you share an example how to use this classes especially the htmx class in my controller with grid please? Then I cant test and give feedback

jpsteil:
If you look at the htmx app I'd shared on the google group, all you'd have to do it change line 159 from:

`grid.param.target_element = "#htmx-target"`

to

`grid.attributes_plugin = AttributesPluginHtmx('#htmx-target')`

AttributesPluginHtmx would have to be imported from grid. (this may be temporary as AttributesPluginHtmx may not stay in grid but be something you'd have to code yourself)

mdipierro:
I like this. Can anybody think of a downside?

mdipierro:
I will just merge this. we can change it if anything comes up. :-)

PR #514, opened 2021-05-04
Fixes to form defaults and Auth defaults

lucadealfaro:
This PR fixes two different problems. 

First, the code for a form did not account for the possibility that the field default is a callable. 
But this is useful to do: 

```
def get_user_email():
    return auth.current_user.get('email') if auth.current_user else None

db.define_table(
    'prova',
    Field('user_email', default=get_user_email),
    Field('name'),
)
```

so that the default for `user_email` is the email of that one particular user.  Pydal respects this, but the form does not.  This is now fixed.  I wonder if there are other related bugs (is the code for grid correct?). 

The other problem this fixes is the defaults for the `__init__` method of `Auth`; some of them had default value set to a mutable, which in Python is dangerous.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -172,11 +172,11 @@ class Auth(Fixture):
         registration_requires_confirmation=True,
         registration_requires_approval=False,
         inject=True,
-        extra_fields=[],
+        extra_fields=None,
         login_expiration_time=3600,  # seconds
-        password_complexity={"entropy": 50},
+        password_complexity=None,
         block_previous_password_num=None,
-        allowed_actions=["all"],
+        allowed_actions=None,
         use_appname_in_redirects=None,
     ):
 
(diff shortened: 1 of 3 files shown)


lucadealfaro:
I see that the build tests fail, but I think this is an issue with the tests; I wonder if Massimo can have a look.

lucadealfaro:
In fact, the tests are expecting people to register even with passwords of low complexity. 
This is incorrect.  If one wants to specify low complexity, one has to lower the entropy by hand.  The default has to be 
to use the standard setting.  So I claim it is the tests that need fixing.  I can fix them if needed.

lucadealfaro:
Let me be more explicit.  Here is why the old declaration: 

```
def __init__(self, ..., 
    password_complexity={'entropy': 50}, ...): 
```

is dangerous.  I could have one app that uses that declaration, and then does: 

```
auth.password_complexity.update(dict(entropy=0))
```

Then, any other app loaded after said app that trusts the default to be 50, would find that in fact there is no entropy requirement at all, because the single default to the Auth parameters has now been changed to `{'entropy': 0}`. 

So I claim that this PR actually fixes a security vulnerability.

Kkeller83:
@jsteil Does this affect grid?

jpsteil:
I don't think it will. The grid doesn't deal with defaults unless it is in a form. So, this should help the grid as well.

PR #517, opened 2021-05-05
GitHub dashboards

ali96343:
These great frontend examples (ported on py4web) may be useful for py4web developers and users

https://github.com/ColorlibHQ/AdminLTE   - 135 MB
https://github.com/themesberg/volt-bootstrap-5-dashboard 30 MB
https://github.com/BulmaTemplates/bulma-templates 10 MB


Kkeller83:
This is awesome but we probaly need to move those into a separate repo
given their size.

ali96343:
I personally prefer the black screen of the cisco-router and the vim-editor, but a modern forend cannot be small. 
We don't have enough disk space ?
Here is a separate repo https://github.com/ali96343/facep4w  - 1GB

Kkeller83:
Super!

Disk space is usually not the problem but if someone wants to download a
framework they just don't expect over 100mb etc I think.

But that is up to Massimo.

I guess most devices from PC to raspberry pi could handle the space
required.

It's just unusual.

Let's see what Massimo thinks

ali96343:
Why in the JS world no one thinks about the size of the application and the size of the nodejs directory ?
Beautiful py4web should have real demo applications,   such as AdminLTE .... etc

mdipierro:
Agree. Feel free to add prominent links from examples and documentation to
your templates

mdipierro:
To clarify my previous comment. I am happy to host these apps on the same webside as the main page but this belongs to another repo and links from the examples app. I can create the repo but since you are doing all the work, I think you should create it and I would use it. ;-)

ali96343:
ok,  This is  a good solution.
Thanks

PR #522, opened 2021-05-07
make watcher a separate service

valq7711:
This need to allow to use watcher with gevent (since asyncio doesn't work with gevent).


mdipierro:
Can you please update the documentation accordingly?

valq7711:
This only changes watcher implementation and doesn't affect on py4web run-options (now `--watch`-option works with gevent as well), so there is nothing  need to be updated in the docs. Maybe there is one point to discuss - now watcher runs as subprocess and communicates with py4web via `http://py4web_cli/<cmd>` - endpoint, which is exposed using bottle.route.

mdipierro:
Please give me some time to evaluate any security implication of adding this endpoint.

valq7711:
It is protected  by `cli_key` which should be passed in a custom header `X-Py4web-Cli-Key`

Kkeller83:
Is this also something to consider? https://github.com/gfmio/asyncio-gevent

valq7711:
Some clarifications:
This is very simple `HTTPCLI` class that provides low-level cli-interface to py4web (mainly for internal needs):
https://github.com/web2py/py4web/blob/429e12b9f071f849736051b1c3f985df3c9d83a5/py4web/core.py#L1273-L1312

`HTTPCLI.command` is a decorator, so you can anywhere:
```python
from py4web.core import HTTPCLI
    
@HTTPCLI.command
def echo(msg):
    return msg
```

`HTTPCLI.mount()` - actually  mounts endpoint 

`HTTPCLI.cli` - command processor that maps `<cmd>` to the function (which was exposed by `@HTTPCLI.command`)

The endpoint is protected by `HTTPCLI.cli_key`
https://github.com/web2py/py4web/blob/429e12b9f071f849736051b1c3f985df3c9d83a5/py4web/core.py#L1286-L1290

Usage example:

https://github.com/web2py/py4web/blob/429e12b9f071f849736051b1c3f985df3c9d83a5/py4web/watcher.py#L31-L42

valq7711:
@Kkeller83 


> Is this also something to consider? https://github.com/gfmio/asyncio-gevent

I tried this first  - doesn't work on windows at least (import error  - it seems that it assumes asyncio old version)

mdipierro:
this is good code but I feel it adds complexity that is not necessary and also exposes a service that one many now want exposed.

PR #525, opened 2021-05-09
update widget support to pass field and vars - add wrapped widgets to…

jpsteil:
Both items up for discussion - maybe there is a better implementation

Widget support - change the parms that are passed to widgets to table and value to field and vars - now in a widget you can reference the field (from which you can get the table) and all vars for the row/record in question.  This gives more flexibility to building a widget. 

Here is a sample radio widget
```
def radio_widget(field, values_dict):
    control = DIV()
    if "_table" in dir(field):
        tablename = field._table
    else:
        tablename = "no_table"
    for k, v in get_options(field.requires)[1:]:
        attrs = {}
        if values_dict and str(values_dict["order_type"]) == k:
            attrs["_checked"] = "ON"
        control.append(
            INPUT(
                v,
                _id="%s_%s" % (tablename, k),
                _value=k,
                _label=v,
                _name=field.name,
                _type="radio",
                **attrs,
            )
        )
        control.append(LABEL(v, _for="%s_%s" % (tablename, k)))

    return control
```

Wrapped Widgets - in custom forms, the DIV wrapper is not included.  This is a good thing. However, sometimes you want that outer div included so you don't have to code it all the time in your custom forms.  This PR adds a 'wrapped_widget' element to the form.custom dict. Then you can choose where this needs to be included. 

This is then used when building custom search forms in grid.py.

Please let me know what issues you see with this and how we can make it better.  In the case of wrapped_widgets - please let me know if you think it's a bad idea.


mdipierro:
It is unclear to me why we need both widgets and wrapped widgets

jpsteil:
I don't want to mess with the behavior of 'widgets'.

When you build a custom form, the outer div from form is not included in form.custom['widgets'].  For some css frameworks you need the outer div to properly format a control (in my case, a bulma select control). In grid, when building a custom search_form that the developer passes in we build a custom form. Without the outer div, the bulma select control (and there are other css frameworks that have the same issues) the control is not rendered correctly.

This change adds the ability for the developer to choose whether or not that outer div is included in the custom form.  I use custom forms heavily and am always having to selectively code this outer div when I know that form.py already has the mechanisms to do it.  

This takes advantage of the formstyle in form.py to decide whether or not an outer div is needed by a specific control.

Let me know if you'd like to see some examples and I can work on that this week.

jpsteil:
Should this get approved I will work on the documentation to note the change and why both are present.

Kkeller83:
I think it makes a lot of sense. Sometimes you will like to have the outer DIV with the widget and sometimes you dont depending on on the framework. If you do using this is convinient. I also chatted with Jim and after this is merged we will start organzing an documentation sprint asap which will explain this.

jpsteil:
This is all addressed in Massimo's new forms.  I am closing.

PR #528, opened 2021-05-10
classes, styles and actionbuttons are now callable in a grid

kevinkosterr:
with this change, the classes, styles and action buttons are now callable.

example usage:
```python
@action("example/<path:path>")
def example(path=None):

    pre_action_buttons = [
        lambda row: ActionButton(
            URL("test", row.id),
            text="Click me",
            icon="fa-plus",
            additional_classes=row.id,
            additional_styles=["height: 10px" if row.bar else None],
        )
    ]

    post_action_buttons = [
        lambda row: ActionButton(
            URL("test", row.id),
            text="Click me!!!",
            icon="fa-plus",
            additional_classes=row.id,
            additional_styles=["height: 10px" if row.bar else None],
        )
    ]

    grid = Grid(
        path=path,
        query=db.foo,
        pre_action_buttons=pre_action_buttons,
        post_action_buttons=post_action_buttons,
    )

    return dict(grid=grid.render())
```

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -1006,6 +1006,8 @@ class Grid:
                                 # if None was returned, no button is available for this row: ignore this value in the
                                 # list
                                 continue
+                        if btn.onclick:
+                            btn.url = None
                         td.append(
                             self.render_action_button(
                                 btn.url,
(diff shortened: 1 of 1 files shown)


kevinkosterr:
> This seems really useful in providing exactly the button you want without the py4web grid dictating so much of it. I think this would also be useful to implement on the CRUD buttons?

This was indeed an oversight, will change it now. Thanks!

PR #539, opened 2021-05-22
Fixes to allow for gcloud deployment

lucadealfaro:
So, it turns out the idea of installing modules in gae/lib/ is problematic. 
The problem is that if there are modules that are BINARY, the binary installed on the local computer is not what is needed in the google cloud.  This was breaking in particular mysqlclient , which was installed in binary on my Mac. 
I think in the end the use of requirements.txt is safer; in this way the cloud uses whatever module is appropriate for it without us trying to guess. 

I have fixed the installation procedure and updated the gae/README.md


mdipierro:
Thank you. That procedure was designed to avoid gevent what it was
required. I agree makes no sense any more

Issue #541, opened 2021-05-23
Validator IS_IN_SET leads to error for "list:"-type fields when form is processed


anonymous378:
models.py:
```
b.define_table(
    'colors',
    Field('colors_available', type="list:string", default=["blue"]),
    )

# This line leads to an error when form is processed and a single color has been selected
# py4web version 1.20210522.2
# py4web/utils/form.py, line 735
# json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
db.colors.colors_available.requires = IS_IN_SET(["red", "blue", "green"])
```
controllers.py:
```
@unauthenticated("colors/add", "colors-add.html", method=["GET", "POST"])
def colors_add():
    form = Form(db.colors, csrf_session=session)
    if form.accepted:
        redirect(URL('index'))    
    return dict(form=form)
```
colors-add.html:
```
[[extend 'layout.html']]
<h1>Colors Add</h1>
[[=form]]
```
This line in py4web/utils/form.py triggers the error when a single string value is selected for the field and the form is submitted:
`original_value = json.loads(original_value or "[]")`

PR #542, opened 2021-05-23
Bug fix for form.py to handle IS_IN_SET validation

anonymous378:
Resolves https://github.com/web2py/py4web/issues/541

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -731,7 +731,10 @@ class Form(object):
                             if field.type.startswith("list:") and isinstance(
                                 original_value, str
                             ):
-                                original_value = json.loads(original_value or "[]")
+                                try:
+                                    original_value = json.loads(original_value or "[]")
+                                except json.decoder.JSONDecodeError:
+                                    original_value = json.loads('"%s"' % original_value)
                             (value, error) = field.validate(original_value, record_id)
                             if field.type == "password" and record_id and value is None:
                                 continue


mdipierro:
Can you post an example to reproduce the problem or the steps? if the field is of type list:something than the value should always be a string containing a list and never a single value. The proposed solution work but, if there is any code path that indeed allows a string which does not contain a list, than the proposed solution would do the wrong this if the value if already quoted as in original_value='"hello world"'.

Kkeller83:
There is some sample code posted in the hug report:

models.py:

b.define_table(
    'colors',
    Field('colors_available', type="list:string", default=["blue"]),
    )

# This line leads to an error when form is processed and a single color has been selected
# py4web version 1.20210522.2
# py4web/utils/form.py, line 735
# json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
db.colors.colors_available.requires = IS_IN_SET(["red", "blue", "green"])
controllers.py:

@unauthenticated("colors/add", "colors-add.html", method=["GET", "POST"])
def colors_add():
    form = Form(db.colors, csrf_session=session)
    if form.accepted:
        redirect(URL('index'))    
    return dict(form=form)
colors-add.html:

[[extend 'layout.html']]
<h1>Colors Add</h1>
[[=form]]
This line in py4web/utils/form.py triggers the error when a single string value is selected for the field and the form is submitted:
original_value = json.loads(original_value or "[]")

anonymous378:
> if the field is of type list:something than the value should always be a string containing a list and never a single value.

"Should be", perhaps, but not what actually occurs.  Look at the difference in how the data gets POSTed in the request.

When the field is specified `Field('colors_available', type="list:string", default=["blue"])` without the requires attribute, the HTML is generated as:
```
<ul class="tags-list"><li data-value="blue" data-selected="true">blue</li></ul>
<input class="type-list-string" placeholder="" list="null">
```
and the POST request contains:
```
Content-Disposition: form-data; name="colors_available"
["blue"]
```

But when the line `db.colors.colors_available.requires = IS_IN_SET(["red", "blue", "green"])` is added, the HTML is generated as:
```
<select class="" id="colors_colors_available" multiple="multiple" name="colors_available">
    <option value=""></option>
    <option value="red">red</option>
    <option selected="selected" value="blue">blue</option>
    <option value="green">green</option>
</select>
```
and the resulting POST request contains:
```
Content-Disposition: form-data; name="colors_available"

blue
```
That value leads to the triggering of the json.decoder.JSONDecodeError because it's not a valid JSON string without double-quotation marks as delimiters.

mdipierro:
Posted a very similar solution based on yours. Thank you.

PR #548, opened 2021-05-31
Switch from YATL to Renoir (#547)

### LINKED ISSUE #547: RFC - Use Renoir templating engine in py4web/yatl
Hi all,
quite a long time since my last commit in web2py repos. Here is a change proposal to use [Renoir](https://github.com/emmett-framework/renoir) templating engine in place of the actual implementation (YATL).

### Summary

[Renoir](https://github.com/emmett-framework/renoir) template engine comes from a major rewrite of the web2py templater, outsourced from Emmett, firstly released as an independent project in Jan 2020. It currently supports Python versions from 3.6 to 3.9, it doesn't require any additional dependency and is released under BSD license.

Its syntax is completely compatible with web2py (and thus with YATL/py4web).

### Rationale

The main reason behind this proposal is to re-unify the codebase of the two projects, giving the chance to py4web and its community to benefit from the enhancements introduced in web2py templater in the last ~6 years of Emmett development.

Obviously, the operation will also introduce the indirect benefit of unify issues, bug triage, enhancements introduction and community, giving thus the chance to make the templating language stronger and popular.

### Benefits

- Adopt modern code
- Support and tests for recent Python versions
- Support for [escaping](https://github.com/emmett-framework/renoir/blob/master/docs/quickstart.md#escaping)
- Support for relative paths in `extend`/`include` instructions
- Support for [extensions](https://github.com/emmett-framework/renoir/blob/master/docs/extensions.md#extensions) like [renoir-haml](https://github.com/gi0baro/renoir-haml)

### Risks

- Splitting the community support across different projects

### Benchmarks

I conducted a super-simple benchmark (might be imprecise), using a simple template:

```html
<div class="posts">
    {{ for idx, post in enumerate(posts): }}
    {{ cssclass = 'white' if idx % 2 else 'gray' }}
    <div class="post {{ =cssclass }}">
        <p>{{ =post['text'] }}</p>
    </div>
    {{ pass }}
</div>
```

and the following code:

```python
import time
from yatl import render as yrender
from renoir import Renoir

POSTS = [
    {"text": "foo"},
    {"text": "bar"},
    {"text": "baz"}
]

def yatl():
    with open("template.html", encoding="utf8") as stream:
        data = stream.read()
    t = time.time()
    for _ in range(20_000):
        yrender(data, context={"posts": POSTS})
    print(f"YATL: {time.time() - t}")

def renoir():
    r = Renoir()
    t = time.time()
    for _ in range(20_000):
        r.render("template.html", {"posts": POSTS})
    print(f"Renoir: {time.time() - t}")

yatl()
renoir()
```

which produced the following results:

```
giovanni@Deneb ~/Downloads/yatl-test ❯ python test.py
YATL: 6.142576217651367
Renoir: 0.4085378646850586
```

### Conclusions

Considering all the upper points, and given the community express its interest in the proposal, I can produce a PR containing the code changes.

Let me know!
Ciao

gi0baro:
SSIA

Resolves #547


mdipierro:
Question about Renoir. The previous code (with yatl.render) used to cache templates until they changed on disk. According to my profiling, the time to read was a bottle neck. How does Renoir handle caching of template sources? What happens they change?

gi0baro:
> Question about Renoir. The previous code (with yatl.render) used to cache templates until they changed on disk. According to my profiling, the time to read was a bottle neck. How does Renoir handle caching of template sources? What happens they change?

@mdipierro it's quite similar, with `reload` parameter enabled it uses `stat` to track changes on files (ref https://github.com/emmett-framework/renoir/blob/master/renoir/cache.py). We can improve this making the reload parameter dynamic based on context. Is there any way in py4web to track if is local development server or production?

Also, I saw your commit to put back yatl. Should I write a `render` function that uses Renoir so we have the same implementation for `Template` and manual `render` calls?

Kkeller83:
A comment in the mailing list came up if there is a chance to include the yatl helpers in renoir?

gi0baro:
> A comment in the mailing list came up if there is a chance to include the yatl helpers in renoir?

@Kkeller83 nope, as I stated in the issue https://github.com/web2py/py4web/issues/547#issuecomment-851663553, python to HTML helpers are out of scope for Renoir, and also Emmett has its own (https://github.com/emmett-framework/emmett/blob/master/emmett/html.py), so it's not really a good fit IMHO.

mdipierro:
This PR is replaced by the simpler one https://github.com/web2py/py4web/pull/608 but need help fixing the problem with yatl.helpers mentioned there. Also the Grid examples do not work. Once these problems are solved I think we should merge this.

PR #555, opened 2021-06-08
Some small improvements  to clearning thread local

lucadealfaro:
I reviewed the previous commit https://github.com/web2py/py4web/commit/9d1bd92960bcf8fa05f1b7133b2f8a81a0af149d . 
This is a small additional change, fixing some comments for the session, and erasing also the flash message.
The flash message cannot normally leak, because if there is no @action.uses(flash) there is no injection in the 
template, but it seems better to just delete these thread local things used at the end of a request. 
In fact, one wonders whether one should really implement an object that is a "request local" somehow. A fixture that provides request local objects, and that guarantees their erasing at the end of the request.  It could be injected in other fixtures via the _ _ requires _ _ of other fixtures.  Thoughts?

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -381,9 +381,13 @@ class Flash(Fixture):
         # if we redirect and have a flash message we move it to the session
         if status == 303 and Flash.local.flash:
             response.set_cookie("py4web-flash", json.dumps(Flash.local.flash), path="/")
-            Flash.local.flash = None
         else:
             response.delete_cookie("py4web-flash", path="/")
+        Flash.local.flash = None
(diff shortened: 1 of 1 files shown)


valq7711:
I believe this is a design problem - direct access to a dozen global objects (i.e. fixtures). It would be better to have something like fixture supplier, so you can access fixture  only using supplier, thus it is impossible to get uninitialized fixture with dirty data. In this case you even doesnt need to write all fixtures  in 'action.uses' , since they will be initialized just by touch 'supplier.session'. All this is implemented in xapi

PR #556, opened 2021-06-08
Made the local objects used in Flash, Session, request-local objects 

lucadealfaro:
I made the thread-local objects used in Flash, Session, request-local in addition to thread local. 
I also added a Current object that can be used to obtain user-level request-local objects. 
Finally, I added a finalize() method to fixtures, that is guaranteed to be called both in case of success or failure; this can be useful (for instance, I use it to ensure that the request-local objects are erased after the request).

### DIFF
--- a/py4web/__init__.py
+++ b/py4web/__init__.py
@@ -20,6 +20,7 @@ from .core import (
     Translator,  # from pluralize
     Session,
     Cache,
+    Current,
     Flash,
     user_in,  # additional fixtures
     URL,  # custom helper
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -351,6 +351,37 @@ thread_safe_pydal_patch()
 # this global object will be used to store their state to restore it for every http request
 ICECUBE = {}
 
+
+#########################################################################################
+# Current Fixture
+#########################################################################################
+
+class Current(Fixture):
+    """
+    This fixture gives access to a request-local object, that is cleaned
+    after each request.
+    """
+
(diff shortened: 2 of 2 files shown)


mdipierro:
Shouldn't Flash and Session use Current? I see they do not have to, but they could extend it. I can make the change if you agree.

lucadealfaro:
It would be a prettier solution, and it would also provide a useful example of how to use Current. 
But there are the following considerations. 

First, should Session, Flash subclass Current?  Probably not.  Current has a dictionary syntax, and Session has a different dictionary semantics, so subclassing is not the right solution here I think. 

So the way to use Current would be to put Current into the `__prerequisites__` of Session.   But then, you would need to create a session as in: 

```
current = Current(),
session = Session(current, ...)
```

and this would break lots of legacy code.  You would then have in Session, 

```
self.current = current
self.current['session'] = object()
```

and instead of `self.local.changed = True`, 

```
self.current['session'].changed = True
```

so I was not so hot on this change.  But perhaps you have a much more elegant way in mind of doing this? 

One might think to have in the Session initializer, 

```
self.current = Current()
```

and then doing simply: 

```
self.current['changed'] = True
```

But this of course would not work, because then there's nothing that would call `self.current.on_request()` at the beginning and `self.current.finalize()` at the end, etc.  The way to chain fixtures is via `__prerequisites__`... 

So I am uncertain.  Perhaps you see a cleaner implementation.

PR #561, opened 2021-06-10
Grid updates

jpsteil:
Need to use wrapped widget to properly format bulma select elements in search form.


jpsteil:
Added another commit so that virtual fields are handled appropriately.  In grid, when a virtual field is in the field list we were still limiting the select statement to only the fields in the list that weren't Virtual fields.  If the field list contains only 1 field and it is a virtual field, then things got messy.  The fix is to not limit the fields returned by the DAL query.  This commit does that.

PR #567, opened 2021-06-16
Make radio buttons look nice and fix colors in auth form

Kkeller83:


agavgavi:
There already is a white class in no.css, so why remake info into another white one when you could change the class used in auth.py around line [158](https://github.com/web2py/py4web/blob/2debec1126baa4e44f1abb6910221dee9a91865a/py4web/utils/auth.py#L158)?

Kkeller83:
Yeah makes sense. I ll change this pr to use the white class instead. What
do you suggest we do about the padding?
Looks really cruddy without it.

Kkeller83:
I'll add thr padding in the html template of the Auth Form

PR #571, opened 2021-06-18
formatters for row add new class and styles

pepej:
To use formatter to row:

grid.formatters['table_name.__row'] = lambda value:  ....

The lambda return   dict(_class='bg-danger',_style= ' ')

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -905,10 +905,18 @@ class Grid:
                 row_id = row["id"]
                 self.use_tablename = False
 
+            key = "%s.%s" % (self.tablename, '__row')
+            if self.formatters.get(key):
+                extra_class = self.formatters.get(key)(row)['_class']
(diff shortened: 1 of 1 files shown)


py4web/utils/grid.py
```diff
@@ -905,10 +905,18 @@ def render_table_body(self):
                 row_id = row["id"]
                 self.use_tablename = False
 
+            key = "%s.%s" % (self.tablename, '__row')
+            if self.formatters.get(key):
+                extra_class = self.formatters.get(key)(row)['_class']
+                extra_style = self.formatters.get(key)(row)['_style']
+            else:
+                extra_class= ''
+                extra_style= '
```
agavgavi, line 914: This looks like it is missing it's second ' in `extra_style= '` vs `extra_style= ''`

mdipierro:
I added a py4web.utils.form.join_class method that should be used in this case

PR #574, opened 2021-06-23
fix FieldVirtual with multiple tables display in grid

jpsteil:
Grid fails if you have a virtual field and a left join (multiple tables).  This code fixes it.

If there is a virtual or custom field in the grid fields then we 

1. Include all param.columns fields that are not virtual or custom
2. Look at each virtual/custom field and add all fields from their table as they might be needed for processing


agavgavi:
Another issue is that the `up`/`dw` variables used for sorting columns in `render_table_header` and `render_field_header` are in the wrong function.  They are both in `render_table_header` when they should be in the new `render_field_header`.

jpsteil:
I just committed a fix for this.  I simply moved the definition to the proper method as it wasn't used in the other one.

agavgavi:
Thanks!

PR #582, opened 2021-06-30
Error in grid if Column doesn't have tablename attribute

jpsteil:
examples/html_grid example wasn't working.  This change fixes that.

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -444,7 +444,9 @@ class Grid:
             # for all FieldVirtual or Custom columns, add all columns from their table
             # as they might be needed for processing
             for col in [
-                x for x in self.param.columns if isinstance(x, (FieldVirtual, Column)) and 'tablename' in dir(x)
+                x
+                for x in self.param.columns
+                if isinstance(x, (FieldVirtual, Column)) and "tablename" in dir(x)
             ]:
                 self.needed_fields.extend(
                     [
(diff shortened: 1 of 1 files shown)


py4web/utils/grid.py
```diff
@@ -444,7 +444,9 @@ def process(self):
             # for all FieldVirtual or Custom columns, add all columns from their table
             # as they might be needed for processing
             for col in [
-                x for x in self.param.columns if isinstance(x, (FieldVirtual, Column))
+                x
+                for x in self.param.columns
+                if isinstance(x, (FieldVirtual, Column)) and "tablename" in dir(x)
```
mdipierro, line 449: I do not think a Column can have a tablename attribute and if it did it may be intended for a different purpose.
I think we should just do:

    for col in self.param.columns
        if isinstance(col, VirtualField):
           ....

jpsteil: I can test that tomorrow with the html_grid example as well as my other grid instances including virtual fields.

mdipierro:
I propose this solution https://github.com/web2py/py4web/pull/583 if you agree feel free to merge

Kkeller83:
Should this be closed in favour of PR 583?

mdipierro:
@Kkeller83 I think so but since one is mine and we have rules, you decide.

Kkeller83:
Ok, cool. Also according to the discussion we had on Discord this can be closed.

PR #593, opened 2021-07-13
Grid improvements

mdipierro:
In this PR:
- removed a vulnerability in Grid that exposed edit/create/delete API even when create=False, editable=False and deletable=False
- allow editable, deletable to be functions of the row
- refactored the per-row button to use the Column logic
- added missing class names for some buttons
- fixed the style of the create button
- fixed the logic behind show_id so that it does not change the field attributes
- various simplifications
- added validation method to constraint what one can create/update
- removed  pre_action_buttons and post_action_buttons because similar logic can be implemented using Column (happy to reinstate if push back from reviewers)
- renamed some internal methods, added "_" prefix for private and replaced "render_" to "make_" for methods that return helpers as opposed to strings.
- added internationalization to grid

still TODO:
- fix inconsistencies in the way classes and styles are applied to different helpers.
- details/edit/create/delete URLs should be signed.
- test with Bulma

### DIFF
--- a/apps/examples/controllers.py
+++ b/apps/examples/controllers.py
@@ -139,7 +139,7 @@ def tagsinput_form():
 # exposed as /examples/htmlgrid
 @action("html_grid")
 @action("html_grid/<path:path>", method=["POST", "GET"])
-@action.uses(session, db, auth, "html_grid.html")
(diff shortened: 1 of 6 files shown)


mdipierro:
resolved merge conflicts

PR #594, opened 2021-07-13
Make get_user() cacheable and auth.user equiuvalent to cached get_user()

mdipierro:
Today:

```
@action(..)
@action.uses(auth.user)
def index():
    return "hello " + auth.get_user()['first_name']
```

with this

```
@action(..)
@action.uses(auth.user)
def index():
    return "hello " + auth.user['first_name']
```

but mind user is a still a fixture and not a dict. Just syntactic sure. Do not feel strongly about this one.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -5,7 +5,7 @@ import datetime
 import hashlib
 import re
 import time
-import urllib
+import threading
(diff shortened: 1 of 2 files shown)


mdipierro:
I recall this and add the caching to the auth.current_user

PR #602, opened 2021-07-16
i18n and p11n UI for dashboard

mdipierro:


mdipierro:
To test. add some T("whatever") to your favorite app.

Kkeller83:
I was testing this today. 

You need to modify e.g. the scaffold app like this:

hello=T("Hello")
greeting=T("Hello {first_name}")

@unauthenticated("index", "index.html")
def index():
    user = auth.get_user()
    message = (greeting.format(**user) if user else hello  )
    return dict(message=message)

To make translations work.

The UI per se is nice, but it loads some entries twice. Then you can not edit the first entry and other smaller UI issues.

I really like the idea of having a UI to manage the translations, but I think we need 2 things: 

1) Just give it a try and you ll see the UI bugs I am talking about instantly. Maybe those can be fixed?
2) Is it worth it fixing if we move to a new dashboard?
3) we need to update the scaffold app with the translation code that works (as written above). 

Happy to create a PR for the scaffold app.

mdipierro:
I tried it. The issue you report are features, not bugs.
What I mean is this:
When you select "update laguages" and it finds T("dog") it always create 3 entries:
0 as in zero dogs
1 as in one dog
2 as in two or more dogs
This may not be the right thing to do maybe it should just create it for counter == 0 or == 1.

It is also true they are not editable. they are deletable and/or repleacible. Again this can be made better with a more complex UI but it accomplishes he goal.

Kkeller83:
Alright lets merge it

mdipierro:
Thanks. We can improve on it later and we should

PR #624, opened 2021-08-30
Allow Auth() to receive kwargs

RichardBH:
Allow Auth() to have kwargs passed in like 

```
mykwargs = {}
mykwargs.update({'format': '%(first_name)s %(last_name)s'})

auth = Auth(session, db, extra_fields= myextra_fields,  kwargs=mykwargs)
```


mdipierro:
I am going to decline this. I see the problem you are trying to solve but this only solved in a particular case and the parameter is a bit obscure.

Currently auth defines a single table but that may not always be the case. Plugins may also define tables. I think it may best to explicitly define the table you want before before creating the Auth instance and you can pass already any parameter you need.

In your specific case you can also do db.auth_user._format="...." after Auth creates the table.

Issue #623, opened 2021-08-29
`finalize` crashes multi-threaded session


valq7711:
https://github.com/web2py/py4web/blob/b68e3565ab77d8536a36acd4d476a77d68749b1f/py4web/core.py#L721-L722

@mdipierro 
It looks like nobody tested this in a multi-threaded environment.
Just in case: `local`-attribute itself is not thread-local attribute;)))))

```
Bottle v0.12.19 server starting up (using RocketServer())...
Listening on http://127.0.0.1:8000/
watching (lazy-mode) python file changes in: D:\_MyPy\py4web\py4web_ombott\py4web\apps
Hit Ctrl-C to quit.

Starting Rocket3 1.3.1
INFO:Rocket.Errors.ThreadPool:Starting Rocket3 1.3.1
Listening on sockets: 127.0.0.1:8000
INFO:Rocket.Errors.ThreadPool:Listening on sockets: 127.0.0.1:8000
127.0.0.1 - "GET /_dashboard HTTP/1.1" - 200 13802
INFO:Rocket.Requests:127.0.0.1 - "GET /_dashboard HTTP/1.1" - 200 13802
127.0.0.1 - "GET /_dashboard/static/js/axios.min.map HTTP/1.1" - 404 459
INFO:Rocket.Requests:127.0.0.1 - "GET /_dashboard/static/js/axios.min.map HTTP/1.1" - 404 459
127.0.0.1 - "GET /_dashboard/info HTTP/1.1" - 200 2311
INFO:Rocket.Requests:127.0.0.1 - "GET /_dashboard/info HTTP/1.1" - 200 2311
127.0.0.1 - "GET /_dashboard/routes HTTP/1.1" - 200 76767
INFO:Rocket.Requests:127.0.0.1 - "GET /_dashboard/routes HTTP/1.1" - 200 76767
127.0.0.1 - "GET /_dashboard/apps HTTP/1.1" - 200 3629
INFO:Rocket.Requests:127.0.0.1 - "GET /_dashboard/apps HTTP/1.1" - 200 3629
ERROR:root:Traceback (most recent call last):
  File "D:\_MyPy\py4web\py4web_ombott\py4web\py4web\core.py", line 867, in wrapper
    [obj.on_success(200) for obj in fixtures]
  File "D:\_MyPy\py4web\py4web_ombott\py4web\py4web\core.py", line 867, in <listcomp>
    [obj.on_success(200) for obj in fixtures]
[...]

PR #628, opened 2021-09-01
Fixture (and Session, Flash) fix: add `_safe_local`-property which is thread safe and reset on every request

valq7711:
Also this PR removes `Current`-fixture.
Reasons:
- it has the same issue as the Session - see #623 
- it is not used by `py4web.core` (there is `py4web.utils` which is better place for such things)


mdipierro:
Thanks Val. This is a better solution.

PR #631, opened 2021-09-06
Agavgavi agavgavi/dashboard (without htmx)

mdipierro:


apps/_dashboard/static/js/dbadmin.js
```diff
@@ -1,10 +1,6 @@
 var app = Q.app();
 var params = new URLSearchParams(window.location.search);
-app.data.loading = 0;
 app.data.app = params.get('app');
 app.data.dbname = params.get('dbname');
 app.data.tablename = params.get('tablename');
-app.data.url = '/_dashboard/rest/{app}/{dbname}/{tablename}'.format(app.data);
-app.data.filter = params.get('filter') || '';
-app.data.order = params.get('order') || '';
 app.start();
```
agavgavi, line 6: I believe this whole file can be removed now that it isn't included anymore.

apps/_dashboard/__init__.py
```diff
@@ -135,10 +201,116 @@ def logout():
...
+            lambda value: value.strftime("%m/%d/%Y") if value else ""
+        )
+
+        grid.formatters_by_type["time"] = (
+            lambda value: value.strftime("%H:%M:%S") if value else ""
+        )
+
+        grid.formatters_by_type["datetime"] = (
+            lambda value: value.isoformat() if value else ""
+        )
```
agavgavi, line 295: These formatters were included to replace the javascript timezone based formatting that already exists in the Grid due to HTMX's issues with injecting Javascript. If we are going back to just normal HTML then I do believe that these are unecessary but I haven't tried it.

mdipierro:
I agree.
Also the css is slightly broken. Would you be able to look into both? If
not I can do it later.

mdipierro:
the lighter the better. I have not tested date/time/datetime

agavgavi:
I am busy tonight sadly with family and unable to do so, my apologies.

mdipierro:
No apology needed. same boat here. Maybe I will find some time later. :-)

mdipierro:
Declining this because we have a new dashboard. We should iterate on that. But thanks this inspired it.

PR #641, opened 2021-10-07
fix Translator fixture, add _has_safe_local property to Fixture class

valq7711:


valq7711:
Right fixtures order does the stuff

Issue #629, opened 2021-09-02
DOCS: The "AUTH UI" paragraph is outdated


nicozanf:
Currently, the [AUTH UI](https://py4web.com/_documentation/static/en/chapter-13.html#auth-ui) is wrong, because the _scaffold/static/components folder is no more present. It seems auth.html is now using full serverside auth forms...

nicozanf:
closed by #658

PR #658, opened 2021-11-24
DOCS: Full rewrite of Authorization chapter

nicozanf:
After sucessfully testing PAM and LDAP (AD) authentication I think we can remove all the warnings ;-)

Also, I've tried to reorganize it and add a couple of pictures to better explain how Auth works technically.


nicozanf:
It closes Issue #629

PR #661, opened 2021-12-05
move hardcoded error messages to MESSAGES dict

d3im:
I'm not sure about edits in AuthAPI - is there dict always available?


jpsteil (changes requested):
instead of using this format 
`self.param.messages["errors"]["invalid_username"]`

can you update to use this format - for consistency with the way the other messages are all accessed

`self.param.messages["errors"].get("invalid_username")`

d3im:
Thanks for revision Jim, it should be done :-)

jpsteil (approved):

PR #662, opened 2021-12-08
DOCS: Grid chapter review

nicozanf:
- rewrite the Company example with the Superheroes, and make it a full working (indipendent) one
- add some screenshots, also with bulma stylesheet

See a preview on  https://nicozanf.github.io/py4web-doc/en/master/chapter-14.html


docs/chapter-14.rst
```diff
@@ -2,102 +2,203 @@
...
+      db.person.insert(name='Bruce Wayne', job='CEO')
+      db.commit()
+
+   @action('index', method=['POST', 'GET'])
+   @action('index/<path:path>', method=['POST', 'GET'])
+   @action.uses(db, 'grid.html')
+   def index(path=None):
+         grid = Grid(path,
+                  formstyle=FormStyleDefault, # FormStyleDefault or FormStyleBulma
+                  grid_class_style=GridClassStyle, #GridClassStyle or GridClassStyleBulma
```
agavgavi, line 59: If this example is supposed to be able to be copy pasted, these two classes should be imported at the top. If it's just a generic example, then this is fine.

nicozanf: You're right, thank you! fixed

PR #675, opened 2022-01-15
new simplified fixture logic

mdipierro:
Experimental branch. All tests passing.
This branch simplifies the syntax of fixtures while allowing more. New fixture have these methods only:

on_request(context)
on_success(context)
on_error(context)

context stores status, output, exception, fixtures, processed
This allowed every on_success to process the output depending on the status and on_error to eat or replace an exception and see what other fixtures were listed and which ones were processed while handling the exception.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -341,32 +341,24 @@ class Fixture:
             )
             return False
 
-    def on_request(self):
+    def on_request(self, context):
         pass  # called when a request arrives
 
-    def on_error(self):
+    def on_error(self, context):
         pass  # called when a request errors
 
-    def on_success(self, status):
+    def on_success(self, context):
         pass  # called when a request is successful
-
-    def finalize(self):
-        pass  # called in any case at the end of a request
-
-    def transform(
-        self, output, shared_data=None
-    ):  # transforms the output, for example to apply template
-        return output
 
 
 _REQUEST_HOOKS.before.add(Fixture.__init_request_ctx__)
 
 
 class Translator(pluralize.Translator, Fixture):
-    def on_request(self):
+    def on_request(self, context):
         self.select(request.headers.get("Accept-Language", "en"))
 
-    def on_success(self, status):
+    def on_success(self, context):
         response.headers["Content-Language"] = self.local.tag
 
 
@@ -374,15 +366,15 @@ class DAL(pydal.DAL, Fixture):
 
     reconnect_on_request = True
 
-    def on_request(self):
+    def on_request(self, context):
         if self.reconnect_on_request:
             self._adapter.reconnect()
         threadsafevariable.ThreadSafeVariable.restore(ICECUBE)
 
-    def on_error(self):
+    def on_error(self, context):
         self.rollback()
 
-    def on_success(self, status):
+    def on_success(self, context):
         self.commit()
 
 
@@ -458,7 +450,7 @@ class Flash(Fixture):
     def local(self):
         return self._safe_local
 
-    def on_request(self):
+    def on_request(self, context):
         self._safe_local = types.SimpleNamespace()
         # when a new request arrives we look for a flash message in the cookie
         flash = request.get_cookie("py4web-flash")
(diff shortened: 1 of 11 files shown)


py4web/core.py
```diff
@@ -872,44 +865,47 @@ def decorator(func):
...
-                    for obj in fixtures:
-                        call(obj, "on_error")
-                    raise
+                    context["status"] = http.status
+                    raise http
+                except Exception as error:
+                    context["exception"] = error
+                    if not isinstance(error, HTTP):
+                        for fixture in reversed(processed):
+                            call(fixture.on_error, context)
```
valq7711, line 902: What if an exception goes here (error in fixture)?

mdipierro: If there is an error in a fixture this results in a ticket. I do not think it is possible for fixtures to handle errors in other fixtures (except on_request).

valq7711: so if we have an exception in fixture before db-fixture, how does db-fixture release the connection?

mdipierro: Good point. Let me fix that.

mdipierro: On second thought. What if there is an error in the fixture that is supposed to release the connection?

py4web/core.py
```diff
@@ -872,44 +864,46 @@ def decorator(func):
...
+                    for fixture in fixtures:
+                        call(fixture.on_request, context)
+                        processed.append(fixture)
+                    context["output"] = func(*args, **kwargs)
                 except HTTP as http:
-                    if http.type == http.Type.success:
-                        for obj in fixtures:
-                            call(obj, "on_success", (http.status,))
+                    context["status"] = http.status
+                    raise http
```
valq7711, line 896: What about fixtures postprocessing?

py4web/core.py
mdipierro: Addressed but I do not think this can be made fool proof.

valq7711: > On second thought. What if there is an error in the fixture that is supposed to release the connection?

I guess It is a bug or total crash. But in case of an exception in before-fixture it could be part of the app-logic

py4web/core.py
mdipierro: we discussed this on the google group. raise HTTP(status, output) is equivalent to response.status=status and return output. So all fixture.on_success are expected to be called except they will see output=="" because the actually output is stored into the HTTP and cannot be modified by the fixtures.

valq7711:
I feel you didn't see all my comments in the google group:

### Layers that have not been punctures should not have anything to say.
> I think of on_request() as a constructor and on_success/on_error as different destructors. If constructor fails is it safe to call the destructor? 

#### Case 1:
Fixture raises HTTP in `on_request` - I suppose it should be appended to `processed`

#### Case 2:
Suppose we have some fixture that uses Auth (it sets some header and needs info about user from Auth), so it should be placed behind, but Auth can raise 403 (i.e. we have logged in user)! and in this case that fixture will be not processed at all.

It seems that guys from Falcon are aware about edge cases like this, as they have an option to process postprocessing-hooks independently from on_request-hooks

mdipierro:
Case 1. yes. it should. we can fix that.

Case 2. auth does not raises 403. auth.user can if the user is missing. If a fixture depends on a user and the user is missing than that fixture in my opinion should not run. We should prevent users from shooting themselves in the foot.

I do not know falcon. If one needs a post-processing hook that is always executed one can create a fixture like

        class Hook(Fixture):
              def post_processing(self, context): do_something()
              def on_success(self, context): return self.post_processing(context)
              def on_error(self, context): return self.post_processing(context)

and one would list this fixture first in uses(hook, ....). This guarantees that post_processing is always called and is always called last. It can access all the other context['fixtures'] and their info. Why is this not sufficient?

valq7711:
> It can access all the other context['fixtures'] and their info. Why is this not sufficient?

it access fixtures after they are destructed

mdipierro:
Fixtures are instantiated when the app starts and are not destructed for as long as the app is running. But you are right in the sense they may have cleaned their thread local state on_success. Can you make a proposal in a PR? I am getting lost.

valq7711:
> Fixtures are instantiated when the app starts and are not destructed

I just use this minds:

> I think of on_request() as a constructor and on_success/on_error as different destructors

I think we need to distinguish wrapper-like fixtures from set-of-hooks fixtures, the latter are like set of event listeners (on_request/on_sucess/on_error) so each hook of these fixtures should be processed independently and guaranteed. 
So we need an attribute like `fixture.is_wrapper`

mdipierro:
It goes against my principle of adding complexity. Now users have to understand the difference between two types of fixtures. I do not even understand it. What is the difference and why is it needed?

mdipierro:
Hello Val,
I added I a commit to this branch which tries to address your concerns without increasing complexity. Now a fixture has a member variable is_hook which defaults to false. If set to true its on_success/on_error methods are always called. WIll this address your concerns?

valq7711:
Hello Massimo!
Yes, it looks fine! (currently I'm diving into a new project at my job, so I don't have free time)

mdipierro:
I am going to wait two more days for additional comments about this. If nobody has concerns I will merge it.

PR #681, opened 2022-01-22
Updates/corrections to auth class.

macneiln:
- Fix to ensure request_reset_password doesn't throw an error if email isn't provided in the request payload.
- Correction to typo variable when specific auth routes are specified rather than 'all'.
- Correction to ensure the ALL_ALLOWED_ACTIONS does not contain duplicated routes (duplicates easily visible with the auth/api/config route).
- Added the missing verify_email AuthAPI endpoint.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -3,6 +3,7 @@ import calendar
 import copy
 import datetime
 import hashlib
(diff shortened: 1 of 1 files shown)


kkellersnow:
This also looks good to me. I think we can merge this.

PR #682, opened 2022-01-30
i18n - grid search can now take utf-8 chars

d3im:
Grid can be searched with strings with accented chars. Maybe too much permissive e.g. emoji or another utf-8 chars and needs revision if not be more restrictive.


d3im:
OK, it must be somehow more restricted since utf chars like 🐝 goes to 500 INTERNAL SERVER ERROR
MySQLdb._exceptions.OperationalError: (1267, "Illegal mix of collations (utf8_general_ci,IMPLICIT) and (utf8mb4_general_ci,COERCIBLE) for operation 'like'")

Issue #445, opened 2021-02-10
Obtain route specific auth.enable() RESTful API Model


macneiln:
How do I get the model for a given RESTful Auth Route API exposed by auth.enable() so that I can build my form client-side before submitting via Axios?

Looking through py4web/auth.py I don't seem to see this as an option other than potentially the "config" auth route but that doesn't really expose the current form field details for each specific route, i.e. login, register, etc. so I'm unsure of it's function.

It would be great if it could function similarly to the PyDAL RESTful API or if there is an existing method please let me know as I don't see anything about this in the documentation either.

*********************
**Chapter 11 Documentation:**
![image](https://user-images.githubusercontent.com/6424306/107474165-be9bff80-6bd6-11eb-9c78-7a7e6cbcc021.png)
*********************
**Chapter 06 Documentation:**
![image](https://user-images.githubusercontent.com/6424306/107474348-10dd2080-6bd7-11eb-980f-6c105e79a4de.png)
*********************

mdipierro:
Sorry do not fully understand. Can you clarify?

macneiln:
Hi Massimo,

I'm hoping that something can be built (if not already existing) for the auth api routes to get the form structure for each route dynamically so the form can be built client side instead of hard coding the form fields in an auth client view. I'm hoping it would work similar (or identical) to the "?@model=true" feature of the RESTful DAL API.

This would return an array of fields with all their attributes, i.e. name, default value, label, type, if the field is readable/writeable, options, regex to satisfy the form field client side, etc. (image above in prior post).

Example:
- Currently to use the **auth/api/login** API you would need to know the form required a user name and password field, whereas I would like to call something like **auth/api/login?@model=true** and have it return the form fields and their attributes so I can dynamically built the form for the user client side. Same with all the auth/api routes such as the **auth/api/register?@model=true**, **auth/api/profile?@model=true**, etc.

That way, it's more generic and flexible if additional fields are added/removed from the auth pages. For example if a decision was made to require address information when registering, the generic client side auth view would not need to be manually updated as these new form fields would be automatically built as long as they are returned from the @model=true view.

mdipierro:
This was done. Thank you!

PR #686, opened 2022-02-13
Auth API - Exposed Model Details

macneiln:
In reference to issue #445, this pull request simplifies some of the auth route exposure logic, as well as implementing a standard method for determining the model relating to exposed Auth API routes.

It functions similar to the [RestAPI](https://py4web.com/_documentation/static/en/chapter-08.html) by returning the API model when parameters **`@model=true`** are sent through in the request. 

**Details:**
- Query parameters **`@model=true`** can be appended to the end of exposed API routes to return the route model for that specific route or the front end can call the "all_models" route to obtain all exposed API route models at once (example below). 
 
**Request Examples:**
- `/auth/api/all_models` **(All exposed models)**
- `/auth/api/register?@model=true` **(Register model)**
- `/auth/api/login?@model=true` **(Login model)**
- `/auth/api/request_reset_password?@model=true` **(Request Reset Password model)**
- `/auth/api/reset_password?@model=true` **(Reset Password model)**
- `/auth/api/change_password?@model=true` **(Change Password model)**
- `/auth/api/profile?@model=true` **(Profile model)**
- `/auth/api/logout?@model=true` **(Logout model)**
- `/auth/api/verify_email?@model=true` **(Verify Email model)**

**Response Examples:**
- Dictionary response potentially containing:
    - **"fields"** array with the fields either visible/editable
    - **"additional_buttons"** array with buttons shown to the user on the corresponding standard auth route
    - **"href"** string for that route
    - **"submit_label"** string with details for that form submission button label
    - **"noform"**  boolean to easily identify routes without forms, i.e. logout/verify_email
    - **"public"**  boolean to determine whether this route should be shown if the user is logged out
    - **"hidden"**  boolean to determine if this is a standard route to shown in a menu list or a hidden route for special situations, i.e. verify_email and reset_password are only shown after they receive an email with a specific URL/token

**Miscellaneous:**
- The current "config" Auth API route should likely be removed as it doesn't seem to be used, or if it used it provides limited value to understanding the exposed routes and their details, which should now be handled but this updated functionality.


macneiln:
**Example Output:**

---
**/auth/api/change_password?@model=true (Change Password model):**

image

---
**/auth/api/login?@model=true (Login model):**

image

---

mdipierro:
wow excellent work. give me a day or two to review.

macneiln:
Sounds good, thanks Massimo.

Kkeller83:
Amazing work!

Issue #687, opened 2022-02-16
500 error message tickets are not created


Kkeller83:
Take this controller:

import os
from py4web import action, Field, DAL
from py4web.utils.grid import Grid, GridClassStyleBulma
from py4web.utils.form import Form, FormStyleBulma
from yatl.helpers import A

# database definition
DB_FOLDER = os.path.join(os.path.dirname(__file__), 'databases')
if not os.path.isdir(DB_FOLDER):
   os.mkdir(DB_FOLDER)
db = DAL('sqlite://storage.sqlite', folder=DB_FOLDER)
db.define_table(
   'person',
   Field('name'),
   Field('job'))

# add example entries in db
if not db(db.person).count():
   db.person.insert(name='Clark Kent', job='Journalist')
   db.person.insert(name='Peter Park', job='Photographer')
   db.person.insert(name='Bruce Wayne', job='CEO')
   db.commit()

@action('index', method=['POST', 'GET'])
@action('index/<path:path>', method=['POST', 'GET'])
@action.uses(db, 'grid.html')
def index(path=None):
      grid = Grid(path,
               formstyle=FormStyleBulma, # FormStyleDefault or FormStyleBulma
               grid_class_style=GridClassStyleBulma, # GridClassStyle or GridClassStyleBulma
               query=(db.person.id > 0),
               orderby=[db.person.name],
               search_queries=[['Search by Name', lambda val: db.person.name.contains(val)]])
               
      return dict(grid=grid)

and this template (grid.html)

[[extend 'layout.html']]

[[=grid.render()]]

and introduce any error in the controller of choice.. e.g. identation error. 

Result: no error ticket will be created, but just a plain white page saying error 500. 
The error is only visible in the terminal.

PR #688, opened 2022-02-16
return error_page instead if setting response.body

valq7711:


valq7711:
fix #687

Kkeller83:
I think it does.

If I remove e.g. an "a" from formstyle=FormStyleBulm instead of formstyle=FormStyleBulma for a form it gives me the 500 error with error ticket now. 

If I have an identantion error I get the red 400 page and see the error message only in the terminal. 

But I think this is how it used to be no?

I am justs not sure anymore for some reason. 

Ideally though the errors that produce the 400 should also create an error ticket no?

Kkeller83:
Thank for the patch this was really cool you fixed it so fast.

valq7711:
@Kkeller83 400? maybe 404  - not found, since indentation error is syntax error and if it is detected at loading time your app is not loaded at all

PR #696, opened 2022-03-12
Scaffold Index missing T usage & Profile Update Issue

macneiln:
Scaffold index missing T usage, erroring when opened.

To recreate: Clone and run py4web locally > open the _scaffold/index route > error occurs due to missing T import.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -1664,7 +1664,7 @@ class DefaultAuthForms:
                         deletable=deletable)
 
         form = Form(
-            fields,
+            self.auth.db.auth_user,
             user,
             formstyle=self.formstyle,
             deletable=deletable,
--- a/apps/_scaffold/controllers.py
+++ b/apps/_scaffold/controllers.py
@@ -31,7 +31,7 @@ from .common import db, session, T, cache, auth, logger, authenticated, unauthen
 
 
 @action("index")
-@action.uses("index.html", auth)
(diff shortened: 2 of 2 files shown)


macneiln:
Commit [7118e0d](https://github.com/web2py/py4web/pull/696/commits/7118e0d2ea11f776c943eba0543b8171dad27709) fixes issue #697, as passing the table fields directly into the Form class sets the dbio flag to false if the value passed to the class is a list rather than the actual table object.

macneiln:
Commit [7118e0d](https://github.com/web2py/py4web/pull/696/commits/7118e0d2ea11f776c943eba0543b8171dad27709) fixes issue #697, as passing the table fields directly into the Form class sets the dbio flag to false if the value passed to the class is a list rather than the actual table object.

py4web/utils/auth.py
```diff
@@ -1664,7 +1664,7 @@ def profile(self, model=False):
                         deletable=deletable)
 
         form = Form(
-            fields,
+            self.auth.db.auth_user,
```
mdipierro, line 1667: why this change?

macneiln: Hi Massimo, this was something I broke with my previous Auth API updates.

The commit fixes issue #697 where the profile details cannot be updated currently due to behaviour in the Form class setting the dbio flag to false if the value passed to the class is a list object rather than the actual table object.

PR #704, opened 2022-03-23
Update chapter-05.rst

AvdN:
The filters `:int`, etc. all have the colon before the ASCII string. `re` can be followed by colon, but the filter name should be consistent with the rest.


nicozanf:
Thank you Anthon!

PR #709, opened 2022-04-02
fixed core

dkawasaka:
fixed core
prevent class name error


mdipierro:
Is this in response to a ticket? Can you help me reproduce the error that requires this fix?

dkawasaka:
> Is this in response to a ticket? Can you help me reproduce the error that requires this fix?

on the dashboard
when I open a table, but not all of them, only some return the error:
type object 'Session' has no attribute '_Fixture__name'

mdipierro:
I am going to try reproduce this first. Something is wrong and the try except will only mask the real problem. Thanks for bringing this up.

mdipierro:
I think this was fixed in a different way.

PR #710, opened 2022-04-02
add groupby option

jpsteil:
Allow groupby on the pydal statement

This requires the use of Column when listing fields.

Please comment with concerns.  This will need doc to explain how to do it since you can't just list the fields without the use of Column().

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -504,6 +504,11 @@ class Grid:
 
         self.record_id = safe_int(parts[1] if len(parts) > 1 else None, default=None)
 
+        table = db[self.tablename]
+        if not self.param.columns:
+            # if no column specified use all fields
+            self.param.columns = [field for field in table if field.readable]
+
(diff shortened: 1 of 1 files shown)


jpsteil:
I've updated this PR to so that you don't have to use Column() class to build your aggregate function columns.  You can now do:

```python
    total_amount = db.agent_transaction.total.sum()
    total_tons = db.agent_transaction.tons.sum()
    grid = Grid(
        path,
        columns=[db.agent_transaction.period_ending, total_tons, total_amount],
        query=query,
        orderby=[~db.agent_transaction.period_ending],
        auto_process=False,
        groupby=[db.agent_transaction.period_ending],
        create=False,
        details=False,
        editable=False,
        deletable=False,
        grid_class_style=GridClassStyleBulma,
        formstyle=FormStyleBulma,
        rows_per_page=10,
        include_action_button_text=False,
    )
```
to groupby and use aggregate functions.

PR #713, opened 2022-04-07
add header and trailer elements

jpsteil:
Allow user to specify additional html elements in above or below the grid.  The py4web grid doesn't allow for export as the web2py grid does.  With the addition of custom html elements above or below the grid, the user can build their own grid exporter functions and add buttons to the grid to call the functions.

Of course you're not limited to export functions.  You can specify any html element.  You can pass a callable or a GridActionButton-like class to build the html element.

header elements are added directly after the New button
footer elements are added on a new row below navigation footer

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -101,6 +101,8 @@ class GridClassStyle:
         "grid-search-form-tr": "grid-search-form-tr",
         "grid-search-form-td": "grid-search-form-td",
         "grid-search-boolean": "grid-search-boolean",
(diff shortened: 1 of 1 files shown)


jpsteil:
updated to allow user to pass html string to be evaluated

added specific classes and styles to handle these elements if you pass GridActionButton-like class

mdipierro:
Since these changes are quite large and add functionality I think we need to make sure of two things:
1) There is not already a way to achieve this
2) The new functionality is documented in the same PR

PR #725, opened 2022-05-06
fix double form submission error

jpsteil:
The form isn't getting signed until after it is processed.  I moved the signing of the form ahead of the processing.

Does this cause other issues?


mdipierro:
Need to test this tomorrow. I am afraid is may introduce a security vulnerability. We should understand when this broke.

mdipierro:
Unfortunately the problem is difference and this create a potential security issues since it will accept all forms. Anyway, pushed a fix.

PR #731, opened 2022-06-02
encrypted session

mdipierro:


lucadealfaro (commented):
This seems to solve the problem... pycryptodome or pycrypto? (see comment).  From https://github.com/pyca/cryptography , cryptography seems easier to use, but perhaps what you are doing with urandom is necessary, I am not quite sure what it does.

PR #736, opened 2022-06-06
Luca components new small fixes

lucadealfaro:
This fixes the deletion of rows in the vue table + forms, and it also removes storage.db from the showcase app (I don't think it should be there, right?).


lucadealfaro:
Whoops.  This is a PR that should be done against a different branch.

PR #742, opened 2022-06-17
Proposed fix for gevent.

remcoboerma:
* aborts when gevent and any form of watch is enabled thus informing the user of what's wrong and what to do about it.
 * adds gevent to server_adapters.py, similar to ombotts version, but removes the reloader option which otherwise breaks the gevent wsgi server.

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -1584,6 +1584,11 @@ def start_server(kwargs):
         params["keyfile"] = kwargs["ssl_key"]
 
     if server_config["server"] == "gevent":
+        if kwargs['watch'] != 'off':
+            print('Error: watch doesn\'t work with gevent. ')
+            print('invoke py4web with `--watch off` or choose another server. ')
+            exit(255)
+
         if not hasattr(_ssl, "sslwrap"):
             _ssl.sslwrap = new_sslwrap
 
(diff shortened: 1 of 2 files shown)


ali96343:
1 this file -  server_adapters.py -  likes long names with Server suffix
probably not gevent , but gevenServer

2 if server_config["server"] == "geventServer":
        if kwargs['watch'] != 'off':
            kwargs['watch'] = 'off'
            print('Info: watch doesn\'t work with gevent. ')
            remove exit(255)

3 it is interesting for me to try geventServer            
   this is a useful patch

remcoboerma:
Hi ali, thank  you for the feedback. 

1) as a function name i'm all okay, but you use this for the cli option as well, and py4web now already mentions `gevent` as a `--server` option. As do [the docs online](https://py4web.com/_documentation/static/en/chapter-03.html#run-command-option), so i propose to stick to the old
> ```
>   -s, --server [default|wsgiref|tornado|gunicorn|gevent|waitress|
>                geventWebSocketServer|wsgirefThreadingServer|rocketServer]
> ```

2) explicit is better than implicit. I've been bitten more than once by a mere warning somewhere is a stream of output when  provided with different functionality than i asked for because of a mix of invalid cli parameters. It's a very efficient way to build developer frustration and disatisfaction. If the user expects both gevent and watch, it is a error, not what warning when the combination is impossible. At least, to me. 

3)  completely agree, gevent would be great to get back. especially with uvloop and all such goodies. #522 would be even better, but requires IPC. This PR is just to have a  go at gevent.

PR #743, opened 2022-07-01
added a http->https redirector for wsgi server

ali96343:
added a redirector http -> https for the wsgirefThreadingServer
the redirector is used together with --ssl_cert options
also, the patch allows you to use both http and https modes

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -36,6 +36,10 @@ import uuid
 import zipfile
 from collections import OrderedDict
 from contextlib import redirect_stdout, redirect_stderr
+import socket
+import ssl
+from datetime import datetime
+
 
 import portalocker
 from watchgod import awatch
@@ -116,9 +120,6 @@ Field = pydal.Field
 request = bottle.request
 response = bottle.response
 abort = bottle.abort
-
-# monkey patching bottle/ombott
-request.query.__class__.get, request.query.__class__.getraw = request.query.__class__.getunicode, request.query.__class__.get
 
 os.environ.update(
     {key: value for key, value in DEFAULTS.items() if key not in os.environ}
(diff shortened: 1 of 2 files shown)


mdipierro:
I think it valuable to have a way to force redirection from http to https. Is there any reason your solution is specific for wsgiref? How about a solution in core that is general and works for all servers?

ali96343:
Of course, it is much better to have a redirector-http->https common to all servers and insert it into core.
I have no reason to use a redirect only for wsgiref.
( I needed a redirector and I inserted it ... that's all  )

ali96343:
The redirector has been moved to core.py 
and can be used by other servers.

The method used is quite universal.

For use with other servers, it is enough to replace
the standard call ssl.wrap_socket( .... ) 
with call SmartSocket(.... )

mdipierro:
Let's think more about this. We definitively want the functionality. Does it belong to the socket level? From the user prospective I would like to be able to "require https" for specific apps or specific actions within the app. That way one does not need to modify the wsgi adapters. What do you think? If we had a reliable way to detect https we could easily turn your code into a Fixture @action.uses(requires_https) for example.

valq7711:
@ali96343 
"about cyclic links"  - it is strange, do you pass `scheme=True` when building URL?

ali96343:
I stopped experimenting with URL and ombott because   another problem was discovered - 
ssl.py does not allow http connections,  the problem occurs before ombott, 
on the socket level .


1 cp -a _scaffold sslapp
2 added to sslapp/controllers.py  (with Val recomendation)
........
`import ombott
@ombott.on_route('/sslapp')
def myhook(hook_path=None):
    print(ombott.request.url)
    raise ombott.HTTPResponse(status='301 Redirect', headers=dict(location='https://google.com'))
`
3 ./py4web.py run apps -s rocketServer  --watch=off --port=8000 --ssl_cert=cert.pem --ssl_key=key.pem
4 (set7-py39) [w3p@sun py4web]$ curl -I http://localhost:8000/sslapp
curl: (56) Recv failure: Connection closed

5 Ombott v0.0.14 server starting up (using <class 'py4web.server_adapters.rocketServer.<locals>.RocketServer'>(reloader=False, certfile='cert.pem', keyfile='key.pem'))...
Listening on https://127.0.0.1:8000/
Hit Ctrl-C to quit.

Starting Rocket3 1.3.2
Listening on sockets: 127.0.0.1:8000*

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/site-packages/rocket3/__init__.py", line 1191, in run
    self.run_app(conn)

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/site-packages/rocket3/__init__.py", line 1681, in run_app
    self.environ = environ = self.build_environ(sock_file, conn)

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/site-packages/rocket3/__init__.py", line 1495, in build_environ
    request = self.read_request_line(sock_file)

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/site-packages/rocket3/__init__.py", line 1240, in read_request_line
    d = sock_file.readline()

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/socket.py", line 704, in readinto
    return self._sock.recv_into(b)

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/ssl.py", line 1241, in recv_into
    return self.read(nbytes, buffer)

  File "/home/w3p/anaconda3/envs/set7-py39/lib/python3.9/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)

ssl.SSLError: [SSL: HTTP_REQUEST] http request (_ssl.c:2633)

Tried to send "500 Server Error" to client but received socket error
Stopping Rocket3 1.3.2

PR #748, opened 2022-07-19
update_or_insert try, except

ch-rigu:
Hi I'v added a try except to overcome errors when the insert fails (errors like: UNIQUE constraint failed: thing.name). 
The try, except avoid a 500 http error on the app when the form is processed.

hope this helps!


mdipierro:
This is an undesirable change of behavior. We expect an exception to be raised because a rollback may be necessary. It is not a good practice to wrap SQL errors into silent try-except. py4web should catch them.

PR #751, opened 2022-08-11
add two factor auth

jpsteil:
Add two factor authentication

docs chapter 13 updated to include documentation on how to use it.  @nicozanf - I only updated the source, I did not process all the changes.

In short, added 2 arguments to Auth()

two_factor_filter - allows you to decide which logins need two factor applied
two_factor_methods - specify methods to use to communicate two factor code to the user

Added messages for 2 factor label and a couple error messages

Added auth.param for two_factor_tries and defaulted to 3

For the processing, I added a new endpoint, auth/two_factor that generates the code, calls the two_factor_methods and displays the verification form.

Other thoughts
In web2py two_factor you can filter the two factor logins in a couple different ways, in py4web I didn't provide anything by default, but you can control with the two_factor_filter parm.  In addition, py4web provides a default notification via email.  I did not provide that, but showed an example in the docs.

Let me know  what I missed or if I'm way off base....

### DIFF
--- a/docs/chapter-13.rst
+++ b/docs/chapter-13.rst
@@ -150,6 +150,95 @@ The second one forces the login if needed:
 Here ``@action.uses(auth.user)`` tells py4web that this action requires
 a logged in user and should redirect to login if no user is logged in.
 
+Two Factor Authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Two factor authentication (or Two-step verification) is a way of improving authentication security. When activated an extra step is added in the login process. In the first step, users are shown the standard username/password form. If they successfully pass this challenge by submitting the correct username and password, and two factor authentication is enabled for the user, the server will present a second form before logging them in.
+
+There are a few Auth settings available to control how two factor authentication works.
+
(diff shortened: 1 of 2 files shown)


mdipierro:
I like this. It is simple and is it clean. In the future the two factor page could be more customizable and ask the use how to send the code (email, phone, etc. assuming the info is in the db). I would like to change the names of the session variables but I can do that myself after merging.

mdipierro:
It would be nice to have an example function for using twilio or ifttt for send an sms. We can send that later

jpsteil:
I'm using sms thru RingCentral. I can include a sample off that in the doc
too.

Is also like to figure out a 'push' verification but don't have a provider
account to test it with.

But yes, I think it will be pretty simple to add more options.

I have no problem changing names of variables. Note, they are mentioned in
the doc and should be updated there as well.

Jim

nicozanf:
Well done @jpsteil & @mdipierro !

We'll process the PDF output of the DOCS on a later stage. I'm planning some other small changes, mainly feedbacks from google group's emails

PR #752, opened 2022-08-17
duplicate query parms fix on pager elements if GET form

jpsteil:
A while back I added the option to have GET verb available for forms.  This seems to have caused some duplicate query parms being created on the pager buttons at the bottom of a grid.  The duplicate query parms weren't causing any problems that I'm aware of, but they shouldn't be there.

this PR attempts to address that.

It is working in my tests.

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -1223,7 +1223,14 @@ class Grid:
         for page_number in self.iter_pages(
             self.current_page_number, self.number_of_pages
         ):
-            pager_query_parms = dict(self.query_parms)
+            #  params can sometimes come through as a list - if so, we only want the last element of the list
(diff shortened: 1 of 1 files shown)


py4web/utils/grid.py
```diff
@@ -1223,7 +1223,14 @@ def _make_table_pager(self):
         for page_number in self.iter_pages(
             self.current_page_number, self.number_of_pages
         ):
-            pager_query_parms = dict(self.query_parms)
+            #  params can sometimes come through as a list - if so, we only want the last element of the list
+            pager_query_parms = dict()
```
mdipierro, line 1227: The [-1:] feels wrong. Can the parameter every be a list? Can you provide an example of how to reproduce?

PR #755, opened 2022-09-07
reset password - clear action token

jpsteil:
- request reset password - this sends an email to the user and adds the token to action_token
- reset password - changes the password but doesn't remove the reset token from action_token

This change updates the change_password() method in Auth to accept a clear_action_token parm which, when True will blank out the action token field.

The problem is that if the reset token is left in the action_token field, the reset email link is still valid and could be used over and over to reset a user password.

The other possible changes I considered were
- adding an Auth.on_accept() method for reset_password that would cleaer the action_token
- adding code to DefaultAuthForms _postprocessing to check if the action is reset_password then we'd clear the action_token

If you prefer this handled differently let me know and I can code it.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -679,7 +679,13 @@ class Auth(Fixture):
     # Methods that assume a user
 
     def change_password(
-        self, user, new_password, old_password=None, check=True, check_old_password=True
+        self,
+        user,
+        new_password,
+        old_password=None,
+        check=True,
+        check_old_password=True,
+        clear_action_token=False,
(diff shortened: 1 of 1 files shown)


jpsteil:
Closing this PR - I've handled by adding the following in my common.py

```
def reset_action_token(form, user):
    auth_user = db.auth_user(user.id)
    auth_user.update_record(action_token=None)
    db.commit()


#  setup on_accept callback for reset_password
auth.on_accept["reset_password"] = lambda form, user: reset_action_token(form, user)
```

PR #778, opened 2023-03-29
Remove doc about IRC py4web channel as it is empty

mkostrikin:
Look like IRC is not actual anymore with discord and other options.
Thus lets drop Documentation referring to IRC channel

Also Cosmetic change for docs/README.md


mkostrikin:
Working on this, I found that `updateDocs.sh` makes a lot of fancy updates including fix for search at doc page.
But, its uses jQuery that suppose to be removed, but maybe it is not the case.

PR #780, opened 2023-03-29
Fix deleting grid records

cdb88:
Records cannot be deleted from the Grid.


cdb88:
I see that there is already a pull request on this topic.

PR #794, opened 2023-05-29
adding logging_level and workers keys to server_adapters

ali96343:
accept -L and -w cmd-keys in server adapters


mdipierro:
What is the use case given that the routes are available via the _dashboard API?

PR #802, opened 2023-07-18
rm server-py4web.log

ali96343:


ali96343:
yes, you are right, there are a lot of unclear issues with the log file for me
- where to put it,
- where to put it in windows
- how to set access rights for /var/log
- novice users definitely do not need a log file

ali96343:
added os.environ.get("TEMPDIR", "/tmp") for log_routes()

ali96343:
added log_dir = os.environ.get('PY4WEB_LOGS', None)

PR #804, opened 2023-07-28
Google Scoped Login

lucadealfaro:
I added a new auth plugin, which implements google Oauth2 login with the ability to get additional scopes. 
There are some differences wrt. other OAuth2 plugins: 
        - The plugin uses the database, and creates an auth_credentials table to 
          store the credentials for the scopes requested. 
        - The plugin relies on some google libraries (see on top), so these 
          need to be installed. 
        - The plugin takes in input a .json credentials file that can be 
          downloaded from Google Cloud when creating the OAuth2 credentials.
This has been tested.

### DIFF
--- a/py4web/utils/auth_plugins/oauth2google_scoped.py
+++ b/py4web/utils/auth_plugins/oauth2google_scoped.py
@@ -1,0 +1,191 @@
+# This is a version of google login that also enables the use of other
+# authenticatio scopes (e.g., Google Drive, etc).  The credentials for the
+# scopes are stored, so that the application can access them and use to
+# operate on the scopes (e.g., create files on Google Drive on behalf of
+# the user).
+# See https://developers.google.com/identity/protocols/oauth2/web-server#python
+
+import calendar
+import json
+import time
+import uuid
+
+import google_auth_oauthlib.flow
+import google.oauth2.credentials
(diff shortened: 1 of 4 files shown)


lucadealfaro:
Note: I also changed auth.py so that if one of these OAuth2 plugins is used, and the login expires, the user is redirected by AuthEnforcer to the plugin URL, and not to the form that asks to login with username/password.  In fact, if a site is using one of these OAuth2 plugins, there is most likely no username/password for the user, and so showing the form is quite confusing.

py4web/utils/auth_plugins/oauth2google_scoped.py
```diff
@@ -0,0 +1,199 @@
...
+                those credentials after logout.  This can obviously
+                generate security concerns.
+        """
+
+        # Local secrets to be able to access.
+        assert secrets_file is not None, "Missing secrets file"
+        self.secrets_file = secrets_file
+        # Scopes for which we ask authorization
+        scopes = scopes or []
+        self.scopes = ["openid",
```
mdipierro, line 60: It is not clear which members are supposed to be accessed publicly and which not, for the latter we should use an _ prefix

mdipierro (changes requested):

PR #808, opened 2023-09-01
sa logging minor improvements

ali96343:


ali96343:
tested with python-3.7.16
and rocket, wsgiref, gevent, tornado, aiohttp

PR #814, opened 2023-10-16
Input sanitization

mrworm:
ANY_OF did the trick.


mdipierro:
I do not think this is necessary but let's dicuss in group if you disagree

PR #818, opened 2023-10-17
Update axios.min.js

mrworm:
Update axios.min.js to 1.5.1
CVE-2023-45857
However this is a huge jump in versions and I have no test cases to verify the functionality. I looked if it's used in the code and was not able to find it.


mdipierro:
It is used in dbadmin and showcase, specifically in the vue components. I think we should delete it from the former and upgrade in the latter.

mdipierro:
axios no longer needed here

PR #821, opened 2023-10-24
adding gunicorn to server_adapters

ali96343:
gunicorn works with py4web


ali96343:
To use gevent monkey.patch_all()  , 
run ./py4web.py run apps -s gunicornGevent ......

and set gevent workers
export GUNICORN_worker_class=gevent
export GUNICORN_worker_class=gunicorn.workers.ggevent.GeventWorker

ali96343:
to be continued

Issue #844, opened 2023-12-09
Dashboard Login Issue


raksingh:
When running a clean installation of py4web, the Dashboard does not accept the password that has been defined.
Refreshing the browser after the first attempt then loads the Dashboard as expected. 

py4web: 1.20231115.1 installed via pip
Python: 3.11.6
Browser: Any (Firefox 121, Opera 105)

mrworm:
I also noticed this quite some time ago. But at that point I wasn't able to find any solution. 
I had better luck today and I have now made a request to add the fix into the master:
[https://github.com/web2py/py4web/pull/846](https://github.com/web2py/py4web/pull/846)

nicozanf:
I think this can be closed now ;-)

PR #846, opened 2023-12-13
https://github.com/web2py/py4web/issues/844

mrworm:
https://github.com/web2py/py4web/issues/844


wakamd:
Having the same Dashboard login issue too

PR #849, opened 2024-01-02
adding ip restrictions to  login process

icodk:
**This PR** contains the following additions to the login process (all implemented in the py4web/utils/auth.py):
1. Checks if client ip address is allowed (contains in a file named: **_host_ip_allow.txt_**)
2. Checks if client ip address is denied  (contains in a file named: **_host_ip_deny.txt_**)
**Notes:**
1. Both files are loaded on py4web start up only.
2. If any of the files is not found the check is not performed.
3. If an IP appears in both files the client is denied.
In addition to these restrictions, in case of a failed login (because of wrong credential ),  the client IP will be  registered in a file name: **_host_ip_blocked.txt_**.
After a predefined number of failures, the client IP will be blocked for predefined number of seconds. Which means that even a a login with correct credentials will be denied until the timeout period is over.
A successful login after the timeout period will cause a removal of  the ip from the  **_host_ip_blocked.txt_**.
**Issues:**
The path to  the **_host_ip_allow.txt_** and the **_host_ip_deny.txt_** files are not specified in the PR. 
Since the **_host_ip_blocked.txt_**  file is created by the framework and not by the user, the path happened to be '/' on an ubuntu installation with the machine-setup.sh script from the py4web\deployment_tools\ubuntu

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -8,6 +8,7 @@ import re
 import time
 import urllib
 import uuid
+import json
 
 from pydal.validators import (
     CRYPT,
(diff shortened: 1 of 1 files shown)


py4web/utils/auth.py
```diff
@@ -36,7 +37,75 @@
...
+        blocked_ip = host_ip_blocked[client_ip]
+    else: # start new registration with timestamp in seconds
+        host_ip_blocked[client_ip]={'count':0,'last_failed':get_timestamp()}
+        blocked_ip = host_ip_blocked[client_ip]
+    blocked_ip['last_failed']=get_timestamp()   # update with the last failed time
+    print(request.path)
+    if blocked_ip['count']< BLCOKED_IP_MAX_COUNT:
+        blocked_ip['count']+=1
+        with open(FILE_BLCOKED_IP_COUNT, "w") as fp:
+            json.dump(host_ip_blocked, fp)
```
mdipierro, line 82: unfortunately this is not thread safe.

py4web/utils/auth.py
```diff
@@ -36,7 +37,75 @@
 [ ] Lock account after x failed login attempts.
 [ ] Force new password every x days.
 """
-
+BLCOKED_IP_MAX_COUNT=5
+# timeout for blocked IP: if BLOCKED_IP_TIMEOUT=0 blocking never expired
+BLOCKED_IP_TIMEOUT=5*60         # reenable blocked ip after this time in seconds
+FILE_BLCOKED_IP_COUNT="host_ip_blocked.txt"
```
mdipierro, line 43: Look into core.py for DatabaseErrorLogger. It creates a global database .service/service.storage. Perhaps make auth create a new "web2py_managed_ips" in there with fields "ip, allowed, blocked, counter" and use those.

py4web/utils/auth.py
```diff
@@ -36,7 +37,75 @@
...
+FILE_HOST_IP_DENY="host_ip_deny.txt"
+FILE_HOST_IP_ALLOW="host_ip_allow.txt"
+host_ip_allow=[]
+host_ip_deny=[]
+host_ip_blocked={}
+try:
+    with open(FILE_HOST_IP_ALLOW, "r") as fp:
+         host_ip_allow = json.load(fp)
+except:
```
mdipierro, line 52: pylint will not like this

py4web/utils/auth.py
```diff
@@ -36,7 +37,75 @@
...
+        with open(FILE_BLCOKED_IP_COUNT, "w") as fp:
+            json.dump(host_ip_blocked, fp)
+
+ # called after successfule credentials validation to check if not blocked on too many attempts
+ # or black or white listed
+def ip_allow():
+    client_ip=request.remote_addr
+    is_this_ip_allowd=True
+    if host_ip_blocked:
+        blocked_ip =host_ip_blocked[client_ip]
```
mdipierro, line 90: This is nice. Even if reading from DB make sure all blocked ips are read at statup to minimize disk IO at every request.

mdipierro:
I have not forgotten about this. On my TODO list. :-)

PR #851, opened 2024-01-02
Fix to Google Scoped Login and launch.json

lucadealfaro:
Two fixes: 

launch.json now is set up so that requests are logged to the console, which is very useful in debugging. 

Also, the AuthEnforcer for the Google Scoped Login now checks that the refresh token is present.  This is a mysterious problem, but Google sometimes fails to set the refresh token, in which case credentials refresh will fail.  Now we detect when the token is missing, and we try to enforce logging in again by removing an old (and supposedly deprecated but apparently not) google cookie.


py4web/utils/auth_plugins/oauth2google_scoped.py
```diff
@@ -38,22 +39,38 @@ def __init__(self, auth, condition=None, error_page=None):
...
+            URL(
+                self.error_page,
+                vars=dict(next=redirect_next),
+                use_appname=self.auth.param.use_appname_in_redirects,
             )
+        )
+            
+    def on_request(self, context):
+        super().on_request(context)
+        user = self.auth.session.get("user")
```
mdipierro, line 66: would not rely on user["email"]. instead assume you have self.auth.user_id and user_info = self.db.auth_credentials[self.auth.user_id] if self.auth.user_id else {}

lucadealfaro: Do not merge it. I will submit a better one.

lucadealfaro: The fix to launch.json os good. But the fix to the scoped login is not a good idea. It worked better before. 

The fundamental problem with Google scoped login is that Google often does not include a refresh token on the callback, and so the credentials to access whatever scopes were requested often expire requiring a new login. 

Checking for the presence of a refresh token is a bad idea since sometimes Google never includes one. 

More generally I am undecided whether the Google scoped login code should be removed from py4web. It does work and I use it in notebookgrader, but perhaps the system is not stable enough and should not be part of the py4web codebase. @mdipierro please advise.

lucadealfaro: I can either leave Google scoped login as it was before, or I can remove it.

mdipierro:
if it works, keep it as it was.

PR #870, opened 2024-05-17
Expose example's db to the dashboard

nicozanf:
I'm not sure this is secure


nicozanf:
And test failed ;-(

mdipierro:
It was intentionally not exposed. Exposing the dashboard itself is tricky.

nicozanf:
I suspected it was not secure... I'll add a note in the docs

PR #880, opened 2024-06-09
Translate chapter it

shanayamalik:


shanayamalik:
Italian translation for docs chapters 6-10 in Italian

PR #893, opened 2024-07-13
grid refactoring

mdipierro:
- added support for Expressions in columns
- added support for virtual fields in joins
- _make_field was unused so was eliminated
- eliminated self.formatters (was unused and better handled in Column definition
- changed meaning of self.formatters_by_type. the type key is no longer the Field type but the actual python type of the value to be displayed. In this way it does not conflict with field represent.
- made more consistent use of - and . in class names.
- added a class grid-cell-{name}

This now works for example:

```
db.define_table(
    'thing',
    Field('name', requires=IS_NOT_EMPTY()),
    Field.Virtual('name2', lambda row: row["name"]),
    Field('in_stock', 'boolean', default=True),
    Field('f_date', 'date', default=datetime.date.today()),
    Field('f_datetime', 'datetime', default=datetime.datetime.now()),
    Field('f_time', 'time', default="12:30"),
    Field('properties', 'list:string'),
    Field('image', 'upload', download_url=lambda name: URL('download', name)),
)

db.define_table(
    'part',
    Field('thing_id', "reference thing"),
    Field('name', requires=IS_NOT_EMPTY()),
    Field('weight', "float"),
)

@action('index')
@action('index/{path:path}')
@action.uses('generic.html', db)
def _(path=None):
    query = db.thing.id > 0
    weight = db.part.weight.sum()
    left = db.part.on(db.thing.id==db.part.thing_id)
    grid = Grid(path, query=query, left=left, groupby=db.thing.id, 
                columns=[
                    db.thing.name,
                    db.thing.name2,
                    db.thing.in_stock,  # shows as checkbox, was broken, use unicode now
                    db.thing.f_date,
                    db.thing.f_datetime,
                    db.thing.f_time,
                    db.thing.properties,
                    weight, # did not work before
                    Column("weight", lambda row: row[weight], required_fields=[weight]) # same as above
                ])
    return locals()
```

### DIFF
--- a/apps/showcase/examples/example_html_grid.py
+++ b/apps/showcase/examples/example_html_grid.py
@@ -50,8 +50,8 @@ def example_html_grid(path=None):
         **grid_param
     )
 
-    grid.formatters["thing.color"] = lambda color: I(
-        _class="fa fa-circle", _style="color:" + color
+    grid.columns[3].represent = lambda row: I(
+        _class="fa fa-circle", _style="color:" + row.color
     )
 
     return dict(grid=grid)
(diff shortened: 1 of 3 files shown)


mdipierro:
Got no feedback except on mailing list. Approving my own PR. :-(

Issue #910, opened 2024-08-19
Auth actions always generate links with appname included


kszys:
I run into an issue when trying to migrate my web2py projects into py4web. py4web does not have the `routes.py` which I used befoe to map URLs to applications. I would like to achive the following setup:

```
http://mydomain1.com -> /myapp1
http://mydomain2.com -> /myapp2
...
```

Lacking `routes.py`, I use nginx as a proxy with something like this (example for `mydomain1.com` -> `/myapp1`):

```
server {
    listen 80;

    server_name mydomain1.com;

    location / {
        proxy_http_version 1.1;
        proxy_set_header Host $host;
        proxy_pass http://127.0.0.1:8000/myapp1$request_uri;
    }
}
```

I use the following parameter for the Auth, which should help in not using the `appname` in the redirects to Auth actions:

```
auth.param.use_appname_in_redirects = False
```

For the clarity of following examples, I also use:

```
auth.enable(route="config/auth")
```

This way, the redirect to Auth is done correctly - e.g., when I try to access:

```
http://mydomain1.com/config
```

which requires authentication, I get correctly redirected to:

```
http://mydomain1.com/config/auth/login
```

However, the `next` parameter is configured incorrectly:

```
?next=/myapp1/config
```

Also, the submit button action is set as:

```
action="//mydomain1.com/myapp1/config/auth/login?next=/myapp1/config/"
```

Notice the two `myapp1` in the URL. All this means that Auth uses the `appname` in all these constructed links. I think it should follow the same logic (and use the same parameter) as for the redirects, actually produce something like this expected result:

```
action="//mydomain1.com/config/auth/login?next=/config/"
```

I looked at the Auth code and I figured out how to fix the `next` parameter. It only requires replacing:

```
redirect_next = request.fullpath
```

with the condition:

```
if self.auth.param.use_appname_in_redirects == False:
    redirect_next = request.fullpath.replace(f"/{request.app_name}", "")
else:
    redirect_next = request.fullpath
```

In two functions of `AuthEnforcer` - `abort_or_redirect` and `goto_login`.

On the other hand, fixing the Submit button in the auth form seems to exceed my skills a bit. It is part of how the Auth APIs are exposed and they all use the `appname` in their paths by defualt. 

Also, after some more looking into the code, there may be also other places, when changes would be required - there are number of places where URLs are constructed. In all those places it should be fixed by:

```
URL( ..., use_appname=self.param.use_appname_in_redirects)
```

(or something similar).

[...]

PR #911, opened 2024-08-19
Add `use_appname` parameter to the Form, Grid, and Auth classes

kszys:
Similarly to issues ancountered with URL generation including the `appname` as discussed in #910 , there is a related issue in Grid class. Grid generates several internal links and until now there are no options to avoid having the `appname` in those URLs. 

The PR propose here adds a parameter `use_appname` to the Grid signature (sorry Massimo!) and then modifies the internal logic inside the Grid class to use it, when generating internal links. This allows to avoid putting the `appname` in all the URLs, if not desired. 

The use case is well described in #910 , so I will not repeat it here. 

I tested the changes, and everything seems to be working as expected, but of course a second look would be very useful. Also, possibly the same functionality may be achieved in a different way.

### DIFF
--- a/py4web/utils/auth.py
+++ b/py4web/utils/auth.py
@@ -69,7 +69,10 @@ class AuthEnforcer(Fixture):
 
         if "text/html" not in request.headers.get("accept", ""):
             raise HTTP(403, body={"message": message})
-        redirect_next = request.fullpath
+        if self.auth.param.use_appname_in_redirects == False:
+            redirect_next = request.fullpath.replace(f"/{request.app_name}", "")
+        else:
+            redirect_next = request.fullpath
         if request.query_string:
             redirect_next = redirect_next + "?{}".format(request.query_string)
         self.auth.flash.set(message)
@@ -91,7 +94,10 @@ class AuthEnforcer(Fixture):
                     request.headers.get("json-redirects", "") != "on"
                 ):
                     raise HTTP(403)
-                redirect_next = request.fullpath
+                if self.auth.param.use_appname_in_redirects == False:
+                    redirect_next = request.fullpath.replace(f"/{request.app_name}", "")
+                else:
+                    redirect_next = request.fullpath
                 if request.query_string:
                     redirect_next = redirect_next + "?{}".format(request.query_string)
                 redirect(
@@ -1482,7 +1488,10 @@ class DefaultAuthForms:
                 self.auth.get_or_delete_existing_unverified_account(email)
         extra_form_fields = self.auth.extra_form_fields.get("register", [])
         fields += extra_form_fields
-        form = Form(fields, submit_value=button_name, formstyle=self.formstyle)
+        form = Form(fields, 
+                    submit_value=button_name, 
+                    formstyle=self.formstyle, 
+                    use_appname=self.auth.param.use_appname_in_redirects)
         user = None
         if form.accepted:
             # notice that here the form is alrealdy validated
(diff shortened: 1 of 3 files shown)


mdipierro:
I understand your solution. I want to run more tests before I can be sure this is the right approach. I have limited computer access in the next few days but will try do it tomorrow.

kszys:
Thanks for the feedback! Take your time - I do not want to introduce any new issues ;) Also - if you have a better/more holistic approach to handling routing between custom domains and py4web apps - I am definitely open for discussion/help. I think this is an important feature for production sites. The `routes.py` in web2py was very useful.

mdipierro:
Sorry if slow but this week I have limited internet. will be a lot more
responsive next week.

kszys:
FYI - I am looking now at some issues with Form, and I think there is the same problem with submiting Forms - the `app_name` is always included in the link. I think this would need to be addressed the same way as in Grid (or in another systematic way). I will try to update this PR to address the Form as well.

mdipierro:
i think this needs to be addressed in URL or the web servet. app should not
need to know

kszys:
I just added the commit to the `form.py` to fix the issue for the submit buttons in forms. It works for me, but...

In Form and Grid there is a lot of manipulation of urls without actually using the URL helper. I presume it _could_ be fixed in URL helper, but then a lot of changes are needed at least in Form and Grid (and possibly in Auth, as described in #910). 🤷

kszys:
Another update 🤷 The auto-generated Forms when using Grid were not following the same logic - ensured now the `use_appname` parameter is passed from the Grid to the auto-generated Forms.

kszys:
So... After the fixes of the Form, I realised I can now fully fix the Auth as well. So I patched it and added to this PR. I believe it is now solving the issue identified in #910 entirely. This PR now appears to comprehensively address this issue across Form, Grid, and Auth utilities. Possibly it requires more testing, as I might not be aware of all the dependencies and possible use cases. 

Massimo - I would appreciate you thoughts on this PR. I think the problem described in #910 and affecting Form, Grid, and Auth is something that is important to enable the use of py4web in production environments. It seems like a likely scenerio, where multiple sites would be served from a single py4web installation. 

The only alternative that I can think of for such a use case is running a separate py4web process for each site as the `_default` app.  In a way it would make it cleaner - not requiring for instance `URL(... use_appname=False)` in every single use of the `URL` helper, but feels like it a bit wasteful and defeats the purpose of py4web supporting multiple applications. This is doable, I guess, but then ideally the documentation should make it clear how such a use case could be handled in production.

mdipierro:
I think the proper way to fix this is to use URL everywhere and add some smartness in URL as opposed to pass use_appname everywhere.

kszys:
So... Massimo - I created a separate PR ( #918 ) to propose changes to the URL helper. If you like the idea, the next step would be to modify Form, Grid, and Auth to always use URL helper when manipulating the URLs. Let me know what you think.

mdipierro:
will review tonight. 👍

kszys:
Assuming the Ombott issue described in #918 is patched, the current URL helper implementation should be sufficient to do what is needed to handle in py4web multi-app deployments with serving multiple domains in a clean way. 

The next stage would be then to fix Form, Grid, and Auth to always use URL helper to manipulate the urls. I will see if, I can propose some PRs for that.

As such, this PR can be closed as not-applicable anymore since the solution using URL helper everywhere will be cleaner and more future-proof.

PR #917, opened 2024-09-02
Update grid.py

salvatj:
corrected    field.type == "id"  
should be   field.name == "id"

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -1390,8 +1390,8 @@ def get_parent(path, parent_field):
             if "parent_id" in kvp:
                 parent_id = kvp.split("parent_id=")[1]
 
-    if parent_id and "&" in parent_id:
-        parent_id = parent_id.split("&")[0]
+    if parent_id and "&" in str(parent_id):
+        parent_id = str(parent_id).split("&")[0]
 
     return parent_id
 
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -360,7 +360,7 @@ class Grid:
         if isinstance(query, query._db.Table):
             query = query._id != None
 
-        if fields and any(field.type == "id" for field in fields):
+        if fields and any(field.name == "id" for field in fields):
             show_id = True
 
         self.path = path


mdipierro:
I think this is correct as field.type=="id" since the id field name is customizable. Perhaps the usage of table.id is wrong and should be table._id to indicate the id field whatever its name is.

When is the problem you are trying to solve?

salvatj:
Hi Massimo,

You may be right
Leave the PR in stand-by and I'll try to reproduce the situation where I
detected this issue.
I emailed Jim before, and he asked me to submit a PR (my 1st time I've ever
done it:  :-)

Saludos,
Jorge Salvat
Tel +34.609.310.100
Av. baile Calbetó Barra 10-C, 2º 1ª
25530 VIELHA  (Spain)

salvatj:
if fields and any(field.type == "id" for field in fields):
                      ^^^^^^^^^^
        AttributeError: 'Column' object has no attribute 'type'
-----
parent_id in some instances cannot be evaluated as a string and I'm performing str() functions against it (Jim found out).

salvatj:
Hi Massimo and Jim,

Could not reproduce the situation, but found the error I got weeks ago:

528     if fields and any(field.type == "id" for field in fields):
                      ^^^^^^^^^^
        AttributeError: 'Column' object has no attribute 'type'

    which I solved changing   type  to  name

528     fields and any(field.name == "id" for field in fields):

Besides that, I just found that another Jim's PR has not been applied yet:
Updated my PR #917  to change from:

    if parent_id and "&" in parent_id:
        parent_id = parent_id.split("&")[0]
to:
    if parent_id and "&" in str(parent_id):
        parent_id = str(parent_id).split("&")[0]

parent_id in some instances cannot be evaluated as a string and I'm
performing str() functions against it.
https://groups.google.com/g/py4web/c/hW598V_zWmk/m/O-HQ_rtPAAAJ

Saludos,
Jorge Salvat

jpsteil:
I think Massimo merged my PR for that recently.

Are you on the latest py4web from GitHub?

Jim

mdipierro:
The fields argument is deprecated. Only columns should be used.
Definitively it was never supposed to be that one passes a column as a
field, while a field can be a column.

I think we should remove the fields attribute and the code below
if fields and any(field.type == "id" for field in fields):
show_id = True

mdipierro:
Also where is get_parent (in grid.py) used? Can it be removed?

jpsteil:
I use get_parent extensively in my htmx code.

If you think it doesn't belong in the grid class, I'll need to move it into
my apps. I use it a lot.

Jim

salvatj:
Just tested new master grid.py and is working ok with your changes.
Should I close this PR or you'll do-it?
Thanks Massimo

mdipierro:
Thank you for confirming!

PR #922, opened 2024-09-21
Fix the way href's are being set in DefaultAuthForms

kszys:
It seems that url manipulation in `DefaultAuthForms` in `auth.py` is missing any kind of consideration for the `route` setup in the `Auth` object. This results in the links failing every time if the `route` is anything but the default `route='auth'`. 

This PR aims at fixing this by replacing static strings with dynamically created urls using the `URL()` helper and taking into consideration the value of the `route` paramter defined in the `Auth` object.


kszys:
Added another commit, as I appear to miss a couple of redirects in the first one - became obvious when I tried to use the 2FA.

PR #925, opened 2024-09-28
Auth TOTP fix

kszys:
This PR proposes to add `two_factor_validate` callback to Auth in order to be able to handle custom two-factor code validation, and in particular handling the case of Time-Based One-Time-Passwords (TOTP). 

Two commits are proposed - one to modify the code, and the other to update the documentation.

### DIFF
--- a/docs/chapter-13.rst
+++ b/docs/chapter-13.rst
@@ -226,38 +226,45 @@ Also notice this method can override the code and return a new one.
 two_factor_validate
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


kszys:
...on a second thought, Massimo - **don't merge this PR into master yet**. 

I think it would be actually useful and _user friendly_ to hide some currently exposed plumbing (such as the session variables, form variable, etc). I just had an idea how to make this _better_ 😜

So let me fix it first, and then we can discuss if needed.

kszys:
Ok - I cleaned-up the code and updated the documentation. I think it is ready now for a review 😄

PR #932, opened 2024-10-17
Fix #931 by using psutil to terminate

### LINKED ISSUE #931: signal.SIGKILL isn't available on Windows
The line 
https://github.com/web2py/py4web/blob/48445773149c6ef1cf03dfa301d68f04ee55d918/py4web/core.py#L1721-L1722
added by
- https://github.com/web2py/py4web/pull/904

causes the following error when running on Windows and hitting CTRL+C, as signal.SIGKILL isn't available on Windows:
```
  File "[REDACTED]\.venv\lib\site-packages\rocket3\__init__.py", line 666, in start
    time.sleep(THREAD_STOP_CHECK_INTERVAL)
  File "[REDACTED]\.venv\lib\site-packages\py4web\core.py", line 1722, in kill_all
    os.kill(os.getpid(), signal.SIGKILL)
AttributeError: module 'signal' has no attribute 'SIGKILL'
 > MainThread > 14:19:04.192
 ```
This error results in being unable to stop py4web using CTRL+C as it just keeps running. 

 See: https://discuss.python.org/t/terminateprocess-via-os-kill-on-windows/30882 for details on os.kill on windows - it seems it would be best not to use it if at all possible.

### PR DESCRIPTION
**Pull request by laundmo**
This fixes
- #931

by using psutil to terminate the process tree. This should avoid any of the issues mentioned in https://discuss.python.org/t/terminateprocess-via-os-kill-on-windows/30882

i'm unsure if this will require extra work on https://github.com/nicozanf/py4web-pyinstaller as psutil includes native code (C)

### DIFF
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -1715,11 +1715,11 @@ def start_server(kwargs):
         main_process = psutil.Process()
         children = main_process.children(recursive=True)
         for child in children:
-            child.terminate()  # friendly termination
+            child.terminate()  # friendly termination (SIGTERM)
         _, still_alive = psutil.wait_procs(children, timeout=2)
         for child in still_alive:
-            child.kill()  # unfriendly termination
-        main_process.terminate()
+            child.kill()  # unfriendly termination (SIGKILL)
+        main_process.kill() # SIGKILL
 
     signal.signal(signal.SIGINT, kill_all)
 
--- a/py4web/core.py
+++ b/py4web/core.py
@@ -18,7 +18,6 @@ import http.cookies
 import importlib.machinery
 import importlib.util
 import inspect
-import io
 import json
 import logging
 import numbers
@@ -36,9 +35,9 @@ import urllib.parse
 import uuid
 import zipfile
 from collections import OrderedDict
-from contextlib import redirect_stderr, redirect_stdout
 
 import portalocker
+import psutil
 from watchgod import awatch
 
 # Optional web servers for speed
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: py4web/core.py [hunk 1]
```diff
@@ -1710,7 +1711,15 @@ def start_server(kwargs):
...
-        os.kill(os.getpid(), signal.SIGKILL)
+        # better way to kill, adapted from: https://stackoverflow.com/a/70565806
+        main_process = psutil.Process()
+        children = main_process.children(recursive=True)
+        for child in children:
+            child.terminate()  # friendly termination
+        _, still_alive = psutil.wait_procs(children, timeout=2)
+        for child in still_alive:
+            child.kill()  # unfriendly termination
+        main_process.terminate()
```
mdipierro, line 1722: does this really work? The issue is with threads that seem to get stuck, not processes?

laundmo: Well, worst case its the same as the previous SIGKILL, but also works on windows.

I haven't noticed any threads sticking around, but i'm also on Windows so its only tested there.

Anyone working on linux could always just test it with `pip uninstall py4web && pip install git+https://github.com/laundmo/py4web`

laundmo: > The issue is with threads that seem to get stuck, not processes?

After further consideration, I think I may have misunderstood originally. You're questioning whether this will stop even if a thread is currently "stuck" doing some long-term work. As a test, I added a `while True:` loop to my index controller, which of course leads to that page just waiting. I can still stop py4web with `CTRL+C` even after this, so it seems that isn't a concern.

mdipierro: I will test this some more. if it works on linux then I trust you it works on windows. ;-)

laundmo: @mdipierro Any update on this? would be nice to switch back from using my fork to the main release

mdipierro: I tested it on Linux and this fails to stop the threads. Can you provide a PR that only affects windows?

laundmo: Could you please elaborate on the exact difference you observed between this PR and the released version which just sends SIGKILL?

Threads can't exist without a process on any OS, so does it actually fail to kill some process?

laundmo:
Closing since https://github.com/web2py/py4web/commit/dce200490ba745b7d8d611cd40942f73a121599f also fixes this

PR #935, opened 2024-10-22
Allow py4web to be run using "python -m py4web"

laundmo:
This PR adds a `__main__.py` file which is a special file representing the command line entry point of a package, used when using the `-m py4web` flag of the python executable.

This allows debugging when installed as pip package, for example in VScode, to be used without creating a extra file, simply using the config
```json
{
  "name": "Python Debugger: Module",
  "type": "debugpy",
  "request": "launch",
  "module": "py4web",
  "args": ["run", "apps", "-D", "--watch", "lazy"]
}
```

Open questions: should the command line functions (click stuff) be moved to this file as well, so that core.py can be the core functionality without command line handling?

### DIFF
--- a/docs/chapter-02.rst
+++ b/docs/chapter-02.rst
@@ -103,25 +103,19 @@ to the vscode ``launch.json`` configuration file. Note that if you're using Wind
 forward slash only, like
 "C:/Users/your_name/py4web/py4web.py".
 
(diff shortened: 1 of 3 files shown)


laundmo:
Just updated the docs as well, to point users towards this being possible now

PR #937, opened 2024-10-27
Update README.rst

dan-carroll:
Simple English grammar correction.


dan-carroll:
A very simple update.

PR #941, opened 2024-11-06
replaced deprecated datetime.utcnow() with datetime.now()

Eddiedigits:
literally the title. 

Checked that .isoformat() still works

### DIFF
--- a/docs/chapter-06.rst
+++ b/docs/chapter-06.rst
@@ -711,7 +711,7 @@ it doesn't exist:
    @action.uses(db)
    def index():
        client_ip = request.environ.get('REMOTE_ADDR')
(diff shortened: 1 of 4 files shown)


mdipierro:
py4web intentionally used utcnow() so that it would use utc time everywhere internally as opposed to server time. Using now() aka server time is problematic. We could use now(timezone.utc). Let me think about this more carefully and get back on this asap.

mdipierro:
Here is my proposal. define in `now=datetime.datetime.now(timezone.utc)` in core.py and use that everywhere. Users who prefer to use a different concept of now can redefine it.

mdipierro:
Closing because now we have utcnow() in core

PR #955, opened 2025-02-24
Allow to check across multiple domains in URL validation

misl6:
The recent commit [0b83a76](https://github.com/web2py/py4web/commit/0b83a76a259f3f8a69b682447aa6d43b3664fce3) introduced the ability to validate generated URLs against a specific domain.

This PR expands that functionality by allowing validation against multiple domains, providing greater flexibility for developers working with multi-domain environments.


mdipierro:
This works but let's discuss on the mailing list if this is what we want. Should we perhaps support instead specify domains per app?

PR #959, opened 2025-03-21
actually use "grid-search-form" from style

laundmo:
Previously, it was not possible to set classes for the `<form>` element of the default search form, even though `grid-search-form` was a classes entry which is set for the various styles.

While that would be a single-line fix, i went ahead and inlined a bunch of assignments which used the same variable name. this makes it less likely mistakes will happen in the future if someone ever touches this code.


mdipierro:
Thank you laundmo

PR #970, opened 2025-04-25
Form add back button, and remove submit button when readonly mode

omlinar:
Add back button to the form when _referrer exists.
submit button will not be displayed when the form is in readonly mode.


mdipierro:
I would prefer not to make the Form logic dependedent on Grid logic. I have fixed this in the grid-refactor in very similar way. Prhaps you can suggest improvements after that s merged.

PR #989, opened 2025-07-20
Add logging support to auth module and improve debug information

nicozanf:
Logging now works fine with auth, you can set its level on the settings.py of the app


mdipierro:
I will merge this but may make some minor changes

PR #991, opened 2025-08-04
Big form widgets refactor to allow user-defined widgets, added number widgets

laundmo:
I was quite unsatisfied with the stats of FormStyleFactory and Form widgets. Especially when it comes to extending/overwriting them, the options were sparse - basically limited to overwriting them per field.

This PR is a pretty big refactor for Widgets and how they work.
- Widgets themselves define what types they apply to, or even more complex matches if no direct type match was found.
- Most Widgets should implement 1-2 methods now, one for an editable version and another for a readonly version
	- The readonly parameter of `widget.make()` wasn't used before, instead a default readonly formatting was applied to basically any field.
- While `Widget.make()` still exists, its primarily called by `form_html` which is what creates the div/label structure used for comments, errors, etc. This allows user-defined widgets to change where in the HTML comments, errors, labels etc. appear.
- The Widget class stores most of the context information and non-html widget structure, and updates it when necessary. Widget implementations call `self.field`, `self.vars` etc. when they need this information. This makes it far easier to work with custom logic compared to all of this logic being in `FormStyleFactory`.
- I've tried to keep compatibility with the `field.widget` and `formstyle.widgets` options, which is done through implementing custom Widget subclasses to handle those.

Since this is a big and somewhat opinionated refactor, I haven't written documentation, exhaustive comments, and tests yet. Before I put that work in I'd want to know whether there's any chance this might be merged.

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -3,11 +3,11 @@ import os
 import time
 import uuid
 from abc import ABC
-from typing import Any, Dict, Optional, Type
+from typing import Any, Dict, Optional, Type, Union
 
 import jwt
 from pydal._compat import to_native
-from pydal.objects import Field, FieldVirtual
+from pydal.objects import Field, FieldVirtual, SQLCustomType
 from yatl.helpers import (
     CAT,
     DIV,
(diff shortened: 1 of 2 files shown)


py4web/utils/form.py
```diff
@@ -65,114 +81,387 @@ def join_classes(*args):
...
-            _type=self.type_map.get(field.type, "text"),
-            _id=to_id(field),
-            _name=field.name,
-            _placeholder=placeholder,
-            _title=title,
-            _readonly=readonly,
+    @classmethod
+    def matches(cls, field: Field) -> bool:
+        "Checks if this widget can be used for the field"
+        return cls.type_name == field.type
```
mdipierro, line 93: Notice field.type is not always a string, if can be a SQLCustomType object

PR #992, opened 2025-08-05
docs: Add Altcha captcha solution to manual

jparga:
This PR adds a new section to the py4web manual (chapter 16) that explains how to integrate a simple, client-side captcha solution using the Altcha library.

The new section includes:

Instructions on how to install the Altcha library.

Example controller code to generate and verify the captcha challenge.

Corresponding template code to display the Altcha widget.

This contribution provides a practical, ready-to-use example for developers who need to add a basic captcha to their py4web forms.

### DIFF
--- a/docs/chapter-16.rst
+++ b/docs/chapter-16.rst
@@ -816,7 +816,6 @@ You need to style the tags. For example:
       line-height: 1.2em;
       margin: 2px;
       cursor: pointer;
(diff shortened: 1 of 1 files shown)


jparga:
I changed the method in the fixture on_request instead of on_success. It avoids crashes between sessions in db. Sorry for the delay.

PR #993, opened 2025-08-11
Enable Form dbio for a list of fields from one table, allow delaying processing

laundmo:
I would like to write forms which act on a subset of fields like this:

```py
return Form([
	db.tbl.field1,
	db.tbl.field2
])
```
but currently, this disabled `dbio` due to a `if isinstance(table, list)` check.

But if all fields have the same table, `dbio` is perfectly safe.

Also, to allow more complex logic including potentially multiple forms on a single page (haven't tested this tho), this makes form processing its own method, analogous to Grid

### DIFF
--- a/py4web/utils/form.py
+++ b/py4web/utils/form.py
@@ -1,5 +1,4 @@
 import copy
-import json
 import os
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


py4web/utils/form.py
```diff
@@ -819,96 +839,100 @@ def __init__(
...
+            self.vars[k] = form_vars[k]
+
+        # We only a process a form if it is POST and the formkey matches (correct formname and crsf)
+        # Notice: we never expose the crsf uuid, we only use to sign the form uuid
+        if self.post_vars:
+            self.submitted = True
+
+        if not readonly and auto_process:
+            self.process()
         if self.csrf_protection:
```
mdipierro, line 861: Should this now always run after process() even if auto_process is False?

laundmo: oh, you're right

PR #999, opened 2025-09-03
Only use Grid.FORMATTERS_BY_TYPE if field.represent is default

laundmo:
Fixes https://github.com/web2py/py4web/commit/b2794823de23c327b6ffa64cc92cb53399f2d71e which completely broke `field.represent` for `reference` fields in `Grid`

This only uses `Grid.FORMATTERS_BY_TYPE` when the user did not set a custom `field.represent` by checking `field.represent` against the default represent functions from pydal.

### DIFF
--- a/py4web/utils/grid.py
+++ b/py4web/utils/grid.py
@@ -8,7 +8,7 @@ import datetime
 import functools
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


laundmo:
closed due to being fixed otherwise

PR #1001, opened 2025-09-22
Reapply "Big form widgets refactor to allow user-defined widgets, added number widgets (#991)"

laundmo:
This reverts commit a1f6238 which itself originally reverted #991.
Fixes:
 - passing a new widget class fails due to missing instantiation logic (reported by Tom Clerckx in the mailing list)
 - Clarifies docs to point out backwards compat for old widgets
 - Clarifies in docs that FormStyleDefault (etc) should be cloned before modifying


laundmo:
You can treat this matching system as me caching in some of the performance budget i've freed up with my threadsafevariable optimizations 😉

PR #1006, opened 2025-10-19
Update index.html

DavidiBeMe:
make the url paths relative so that they work with url_prefixes as well.


DavidiBeMe:
I tested this change with and without using url_prefix and it worked in both cases using rocket3

mdipierro:
This worked because the action was @action("index"). If the actions was @action("test") using the same index.html it would break. Let me think some more about this.

